PR #1, opened 2019-01-23
Update deploy

colearendt:
Moving towards the public APIs released in 1.7.0

- add NEWS.md
- start using naming convention `content_` for `content_create`, `content_upload`, `content_deploy`, etc.
- start using `guid` instead of `app_id`
- add some yaml and github helper functions
- improve `poll_task`
- start adding tests
- rename package to `connectapi`
- start using `glue` in place of `sprintf`. It adds a dependency, but makes things more readable and remedies some pain elsewhere... Open to discussion :smile:

Questions:
- do we like the `content_*` endpoints for autocompletion? What about `content_get`? I left that as `get_content` currently... but that feels strange.
- What about users? Should we have `users_lock`, `users_unlock`, etc.?
- how should we handle Connect versioning? I added a simple check_connect_version function that we could run on `Connect$new`, but not sure how we want to focus this. My imagination is that we should basically just track the latest RSC APIs as best we can (and leave the older versions of the package to keep in sync with older versions of the server).


colearendt:
Punting on most of these questions in pursuit of velocity. 😄 Documenting them for future revision if need be

PR #35, opened 2019-11-27
Pagination

tbradley1013:
This PR would modify the two pagination while loops (one in the `get_apps` R6 method and the other in the `page_cursor` function. 

It does some number checks to see if the current iteration of the while loop would cause the output to exceed the requested limit and modifies the limit to restrict the final query to be only the length of the needed remaining values. 

For example, if a user queries the content on a connect system in the current implementation:

```
library(connectapi)
client <- connect()

client$get_apps(.limit = 30, page_size = 25)
```

Then the result will be of length 50 (or however many apps they have if it is between 25 and 50) rather than the limit requested of 30. 

The new implementation will check that the current length of the output `all` subtracted from the requested limit is not less than the page_size. If it is then the page_size is modified for the final query. Now the above code will return 30 records as requested. 

A similar approach was used for the `page_cursor` function. Although this one required that I use regex to modify the `next` url in the previous query. But the general concept is the same. Now if a user requests 600 usage records, the page_cursor function will return 600 records rather than 1000. 

```
tmp <- client$inst_shiny_usage(limit = 634)

page_cursor(client, tmp, limit = 634)
```

The regex looks for the string "limit=" plus any numbers and replaces it with "limit=" plus whatever the newly calculated limit. I am not sure if that's the best way to go about it  but it seems to work for me. 

Also, do you all have a dummy connection you test these things on? I can test them on the system I am the admin for but don't really want to put my servers output up here when I make reprex?

### DIFF
--- a/R/connect.R
+++ b/R/connect.R
@@ -201,6 +201,8 @@ Connect <- R6::R6Class(
         total = NA,
         clear = FALSE
       )
+      
+      if (.limit < page_size) page_size <- .limit
 
       # handle paging
       prg$tick()
@@ -211,9 +213,13 @@ Connect <- R6::R6Class(
           )
       )
       all <- res$applications
+      all_l <- length(all)
       start <- page_size + 1
       while (length(res$applications) > 0 && length(all) < .limit) {
         prg$tick()
+        
+        if ((.limit - all_l) < page_size) page_size <- (.limit - all_l)
+        
         res <- self$GET(
           sprintf(
             '%s%scount=%d&start=%d&cont=%s',
@@ -221,6 +227,7 @@ Connect <- R6::R6Class(
             )
           )
         all <- c(all, res$applications)
+        all_l <- length(all)
         start <- start + page_size
       }
       all
--- a/R/page.R
+++ b/R/page.R
@@ -30,12 +30,12 @@ page_cursor <- function(client, req, limit = Inf) {
     
     # change limit if this iteration would exceed the requested limit
     next_url <- response$paging$`next`
+    response <- client$GET_URL(next_url)
+    
     if ((limit - length(res)) < limit){
       limit <- (limit - length(res))
-      next_url <- gsub("limit=[0-9]*", paste0("limit=", limit), next_url)
+      response$results <- head(response$results, n = limit)
     } 
-    
-    response <- client$GET_URL(next_url)
     res <- c(res, response$results)
   }
   return(res)
--- a/R/page.R
+++ b/R/page.R
@@ -32,12 +32,9 @@ page_cursor <- function(client, req, limit = Inf) {
     next_url <- response$paging$`next`
     response <- client$GET_URL(next_url)
     
-    if ((limit - length(res)) < limit){
-      limit <- (limit - length(res))
-      response$results <- head(response$results, n = limit)
-    } 
     res <- c(res, response$results)
   }
+  res <- head(res, n = limit)
   return(res)
 }
 # TODO: Decide if this `limit = Inf` is helpful or a hack...
--- a/R/page.R
+++ b/R/page.R
@@ -27,7 +27,15 @@ page_cursor <- function(client, req, limit = Inf) {
   res <- response$results
   while(!is.null(response$paging$`next`) && length(res) < limit) {
     prg$tick()
-    response <- client$GET_URL(response$paging$`next`)
+    
+    # change limit if this iteration would exceed the requested limit
+    next_url <- response$paging$`next`
+    if ((limit - length(res)) < limit){
+      limit <- (limit - length(res))
+      next_url <- gsub("limit=[0-9]{1,5}", paste0("limit=", limit), next_url)
+    } 
+    
+    response <- client$GET_URL(next_url)
     res <- c(res, response$results)
   }
   return(res)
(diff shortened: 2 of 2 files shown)


colearendt:
@tbradley1013 Clever solution!! What do you think about just passing the result to a `head()` call right before we return here? Ultimately, we have to make that last API request anyways, and we would be retrieving / throwing away more records than necessary, but that doesn't seem like unnecessary overhead, given that it saves us from having to regex-parse the `next` URL (which admittedly gives me a bit of pause, even though I love regexes!).

The underlying story here is that I'm hoping to keep this paging mechanism a bit generic and not wed too closely to the API's implementation details if possible.

tbradley1013:
I am not opposed to using head rather than the regex. Would you want to use head for both methods or only the `page_cursor` function? I think the implementation for the `get_apps` method is fairly clean but head would work there as well if you would like. 

I ran benchmarks on the regex method and a modified version using `head` and there is very minimal slow down to using the head method. So I can change either just the `page_cursor` function or both depending on your decision. 


``` r
page_cursor <- function(client, req, limit = Inf) {
  qreq <- rlang::enquo(req)
  
  prg <- progress::progress_bar$new(
    format = "downloading page :current (:tick_rate/sec) :elapsedfull",
    total = NA,
    clear = FALSE
  )
  
  prg$tick()
  response <- rlang::eval_tidy(qreq)
  
  res <- response$results
  while(!is.null(response$paging$`next`) && length(res) < limit) {
    prg$tick()
    
    # change limit if this iteration would exceed the requested limit
    next_url <- response$paging$`next`
    if ((limit - length(res)) < limit){
      limit <- (limit - length(res))
      next_url <- gsub("limit=[0-9]*", paste0("limit=", limit), next_url)
    } 
    
    response <- client$GET_URL(next_url)
    res <- c(res, response$results)
  }
  return(res)
}


page_cursor_head <- function(client, req, limit = Inf){
  qreq <- rlang::enquo(req)
  
  prg <- progress::progress_bar$new(
    format = "downloading page :current (:tick_rate/sec) :elapsedfull",
    total = NA,
    clear = FALSE
  )
  
  prg$tick()
  response <- rlang::eval_tidy(qreq)
  
  res <- response$results
  while(!is.null(response$paging$`next`) && length(res) < limit) {
    prg$tick()
    
    # change limit if this iteration would exceed the requested limit
    next_url <- response$paging$`next`
    response <- client$GET_URL(next_url)
    
    if ((limit - length(res)) < limit){
      limit <- (limit - length(res))
      response$results <- head(response$results, n = limit)
    } 
    
    res <- c(res, response$results)
  }
  return(res)
}

client <- connectapi::connect()
#> Defining Connect with host: http://pwdrstudio:3939/
tmp <- client$inst_shiny_usage(limit = 634)


microbenchmark::microbenchmark(
  regex = page_cursor(client, client$inst_shiny_usage(limit = 634), limit = 634),
  head = page_cursor_head(client, client$inst_shiny_usage(limit = 634), limit = 634)
)
#> Unit: milliseconds
#>   expr      min       lq     mean   median       uq      max neval
#>  regex 239.6832 249.1177 257.7384 256.7153 263.9516 323.2866   100
#>   head 248.8592 263.7184 271.4393 268.2116 274.6582 355.2555   100
```

<sup>Created on 2019-12-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/connect.R
```diff
@@ -211,16 +213,21 @@ Connect <- R6::R6Class(
           )
       )
       all <- res$applications
+      all_l <- length(all)
       start <- page_size + 1
       while (length(res$applications) > 0 && length(all) < .limit) {
```
colearendt:
```suggestion
      while (length(res$applications) > 0 && all_l < .limit) {
```

colearendt:
I think we can leave the `get_apps` pattern as you have it! Just one little suggestion since we have the length available... perhaps this is bad practice since it introduces a dependency on state but 🤷‍♂ 

Nice `microbenchmark`! That's a great sanity check, thanks!

tbradley1013:
Looks good. I changed the `page_cursor` to the one shown in the benchmark example. This uses head just on the last iteration rather than on the entire dataset but I don't think that it should make much difference. 

I am not entirely sure what you mean by depending on state? Is that because we are defining all_l before the while loop?

colearendt:
Right, the idea is that if we change `all_l` somewhere or alter its definition, we could subtly break the while loop (and it could run forever / etc.) because the while loop condition now depends on the state of some variable that we have to manually keep in sync with the `all` object.

Sorry to be a stickler here - I think it'd be easier to reason about if we just do the `head` on the end. That way we don't have to think through all the logic gates to determine whether things are working properly. In principle, definitely not much difference. In practice, I think it should be easier to maintain and think through if we just let `head` do the limiting.

tbradley1013:
Sounds good. I moved the `head` outside of the while loop. To be thorough, I ran a similar benchmark and as expected, there is no performance difference between inner/outer:

```
#> Unit: milliseconds
#>        expr      min       lq     mean   median       uq      max neval
#>  head_inner 248.7050 265.9070 274.6828 273.8695 282.7586 315.2592   100
#>  head_outer 249.7441 266.0086 274.9511 272.9677 280.3701 332.6725   100
```

PR #36, opened 2019-11-27
Get functions

tbradley1013:
This PR adds 7 new functions that each query a different endpoint by exposing different R6 methods in the Connect object. 

- `get_users`
- `get_groups`
- `get_group_members`
- `get_content`
- `get_usage_shiny`
- `get_usage_static`
- `get_audit_logs`

Each of these functions simply exposed the differetn arguments of each method. The user needs to simply provide a valid `connectapi::connect()` object. I also added the helper function `parse_connectapi` which is simply the parse code in the `api_build` method. 

Each of these functions has faily thorough documentation detailing the arguments, the output tibble, basic examples, and pointing them to the relevant section in the API reference document. 

```
library(connectapi)
client <- connect()

get_users(client)

groups <- get_groups(client)
groups

get_group_members(client, groups$guid[1])

get_content(client)

get_usage_shiny(client)

get_usage_shiny(client)

get_audit_logs(client)
````

Currently, any of the endpoints that provide paging information (i.e. provide a next value) support paging in these functions.

### CODE REVIEW

**Comment by colearendt** [comment]
This looks awesome! And impressive docs! Did you generate the docs from the API definition / swagger? Or did you copy that over manually? 😱 

I'll try to take a closer look soon!

**Comment by tbradley1013** [comment]
I copy and pasted any argument or output definitions from the API documents where available. So it was a little tedious but didn't take very long once I got into it! But I wanted the descriptions to be consistent across the different places!

**Comment by colearendt** [comment]
Nice! Yes, I definitely agree - consistency is super useful.

**Comment by colearendt** [comment]
One question I had on this one. Is the primary aim of these functions to just save the user the hassle of messing around with the R6 class? Or to provide documentation?

I.e. do you envision that there would be a need for these methods if there was a way to pass all parameters on to the "lazy" tibbles?

A potential pain point I see here is having to maintain parameters in two places for each R6 method. There are also some snazzy new R6 docs patterns in roxygen that we could take advantage of if documentation is your primary aim. Just curious 😄

File: R/get.R [hunk 1]
```diff
@@ -0,0 +1,597 @@
...
+  res <- src$users(
+    page_size = page_size,
+    page_number = page_number,
+    prefix = prefix
+  )
+  
+  ## TODO: Add a pagination option for users? This could be done using the 
+  ## res$total value and changing the page_size and page_number arguments
+  ## with a limit arguemnt. Then a while loop could be written similar 
+  ## to page_cursor
```
colearendt: Yes! I totally agree. I think we will ultimately want to write another `page_` helper that can simplify this pattern / allow for re-use across the different API endpoints that use this type of paging. I haven't had pressing need for this since I have not had need to worry about > 400 users or groups 😅

R/get.R
```diff
@@ -0,0 +1,597 @@
...
+  
+  res <- res$results
+  
+  if (length(res) >= 400) {
+    warning("The 'users' tbl_connect does not page and will return max 500 users")
```
colearendt:
```suggestion
    warning("'get_users' does not page and will return max 500 users")
```

R/get.R
```diff
@@ -0,0 +1,597 @@
...
+  
+  ## TODO: Add a pagination option for groups? This could be done using the 
+  ## res$total value and changing the page_size and page_number arguments
+  ## with a limit arguemnt. Then a while loop could be written similar 
+  ## to page_cursor
+  
+  res <- res$results
+  
+  if (length(res) >= 400) {
+    warning("The 'groups' tbl_connect does not page and will return max 500 users")
```
colearendt:
```suggestion
    warning("'get_groups' does not page and will return max 500 groups")
```

R/get.R
```diff
@@ -0,0 +1,597 @@
...
+#'   client <- connect()
+#'   
+#'   get_content(client, limit = 20)
+#' }
+#' 
+#' @export
+get_content <- function(src, filter = NULL, limit = 25, page_size = 25){
+  validate_R6_class("Connect", src)
+  
+  warn_experimental("'get_content'")
```
colearendt:
```suggestion
  warn_experimental("get_content")
```

I don't think the extra quotes are needed. Just something I did for some weird reason on the `tbl_connect` endpoint 🤷‍♂

R/get.R
```diff
@@ -0,0 +1,597 @@
...
+  
+  return(out)
+}
+
+
+
+
+
+
+parse_connectapi <- function(data){
```
colearendt: Maybe worth pulling this into the `utils.R` file? I think this is something that we will want to use generically elsewhere 😄

tbradley1013:
I think my goal was kind of all of those things. The R6 methods are not documented at all yet (which may be on the todo list) so I had to look at the source code to see what each did. This may be something users aren't exactly comfortable with. In addition, I don't think R6 methods are something that people are use to using so it could be confusing. So I was definitely looking to provide clear docs but mostly it was about making a more standard R interface.

 In addition, the R6 methods don't parse the data before returning it so that leaves it to the user to either dig out your parsing method (which is what I did) or to figure it out on their own (which I think is more likely they will attempt), so these functions provide the parsing so that they get the data in a tibble. 

For the lazy tbls, I like the idea of them, but to me they seem more practical for interactive use of the api. For example, the lazy tbls just require me to pass an additional function (`collect()`) to get the data into my R session or shiny application to be able to build figures while the `get_*` functions allow users to just get the data in one function. Although, once the lazy tbls allow for all of the end point arguments (or at least the ones you want to expose) to be passed to them, I may find them more functional. 

I see you point about the maintenance of the parameters in two places. I totally get it if you decide against these because of that. My thinking was that these functions simply provide a more common API interface similar to the ones seen in other R API interface packages. One potential work around for maintaining it in two places is to just use `src` and `...` as the arguments to all the `get_*` functions and have them reference the R6 method docs once they are done for individual arguments

R/get.R
tbradley1013: Same here. I didn't tackle it because I also don't have a need for it at the moment (and may never). But in theory it  doesn't sound overly difficult since the overall structure can be modeled off the other while loops, so I figured I would leave the comment since it was something I saw you were doing in other places

R/get.R
tbradley1013: I moved it over! I had originally meant to do that but left it here during development and then forgot!

colearendt:
Nice! That makes a lot of sense - thanks for the thorough writeup! I think in the short term, these functions are a no-brainer - definitely much more usable 😄 Long-term, maybe we decide whether the lazy approaches are sufficient or whether a `src`/`...` approach w/ linked R6 docs makes sense (I definitely want to try out [the new snazziness](https://github.com/coolbutuseless/minihtml/blob/master/R/HTMLElement.R) ).

tbradley1013:
Yea. I am happy to help with any changes that need to be made to these in the future as well so feel free to ping me about it. 

I have never used R6 methods myself but they seem super useful and I may try to dive into them in the near future!

PR #37, opened 2019-12-02
Get tags

tbradley1013:
This PR adds several functions to get the tags on the RStudio Connect system. 

- `get_tags` - this just returns the simple two column data.frame returned from the `get_tags` R6 method. 
- `get_tag_info` - this parses the information from the `get_tag_tree` R6 method into a tibble
- `get_tag_tree` - this steals the concept from `fs::dir_tree` and creates a tree in the console of the tags. I have posted a reprex showing what it looks like below. Note that the `get_tag_tree` function only needs the valid connect source object and will create the tree from the servers tags. The reprex just creates the tags via a helper function and runs it through the internal function within `get_tag_tree`. 

Some things to note about the PR which might be issues:
- It adds `stringr` to the Imports section of description. I needed `str_extract`, so I added it. If this is a dealbreaker (i.e. you want to limit your dependencies) then I can rework it with base R
- the `get_tag_info` doesn't utilize the `src$tags`  list but rather the nested list from `src$get_tag_tree`. The helper function seems to work fine but I am not sure whether it would be better to just initiate the creation of `src$tags` and make the parse function a little more robust. I am open to suggestions and will make changes accordingly. 

## Reprex for `get_tag_tree` functionality
``` r
# function to make example tags
make_tags <- function(n, child = NULL){
  purrr::map(1:n, ~{
    name <- randomNames::randomNames(n = 1, which.names = "first")
    out <- list(
      name = name,
      children = NULL
    )
    
    if (!is.null(child)){
      out$children <- make_tags(n = child)
    }
    
    return(out)
  })
}

tmp <- make_tags(2, 2)

tmp[[2]]$children[[2]]$children <- make_tags(3, child = 1)
tmp[[1]]$children[[1]]$children <- make_tags(2, 1)


# functions to make tag_tree
tag_tree <- function(tags, top_tag = "tags"){
  parsed_tags <- parse_tags(tags, top_tag = top_tag)
  
  tag_split <- split(parsed_tags, tag_ancestor(parsed_tags))
  names(tag_split) <- gsub("/$", "", names(tag_split))
  
  ch <- fs:::box_chars()
  
  print_leaf <- function(x, indent) {
    leafs <- tag_split[[x]]
    for (i in seq_along(leafs)) {
      if (i == length(leafs)) {
        cat(indent, fs:::pc(ch$l, ch$h, ch$h, " "), basename(leafs[[i]]), "\n", sep = "")
        print_leaf(leafs[[i]], paste0(indent, "    "))
      } else {
        cat(indent, fs:::pc(ch$j, ch$h, ch$h, " "), basename(leafs[[i]]), "\n", sep = "")
        print_leaf(leafs[[i]], paste0(indent, fs:::pc(ch$v, "   ")))
      }
    }
  }
  
  cat(top_tag(parsed_tags), "\n", sep = "")
  print_leaf(top_tag(parsed_tags), indent = "")
}


parseTags <- function(x){
  parsed_tags <- purrr::map(x, ~{
    out <- .x$name
    # out <- paste(top_tag, out, sep = "/")
    if (length(.x$children) > 0){
      child_names <- parseTags(.x$children)
      child_names <- paste(out, child_names, sep = "/")
      out <- c(out, child_names)
    } 
    
    return(out)
  })
  
  parsed_tags <- purrr::flatten_chr(parsed_tags)
  out <- gsub("/$", "", parsed_tags)
  
  return(out)
}

parse_tags <- function(x, top_tag = "tags"){
  out <- parseTags(x)
  
  out <- paste(top_tag, out, sep = "/")
  return(out)
}

tag_ancestor <- function(x){
  out <- stringr::str_extract(x, "^(.*[\\/])")
  out <- ifelse(is.na(out), paste0(x, "/"), out)
  return(out)
}

top_tag <- function(x){
  unique(stringr::str_extract(x, "^[^/]+"))
}

# print the tag tree
tag_tree(tmp)
#> tags
#> +-- Morgan
#> |   +-- Cherise
#> |   |   +-- Alma
#> |   |   |   \-- Dong-Keun
#> |   |   \-- Rebekah
#> |   |       \-- Omar
#> |   \-- Jessica
#> \-- Michaela
#>     +-- John
#>     \-- Anabella
#>         +-- Jacob
#>         |   \-- Briana
#>         +-- Tyler
#>         |   \-- Damien
#>         \-- Taylor
#>             \-- Antonio
```

<sup>Created on 2019-12-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


colearendt:
Really impressive print formatting for the tag tree! That's awesome!

I have been trying to keep `stringr` out of being a formal dependency as much as possible, since `stringi` is notoriously painful for users to install in offline environments. It is definitely common enough that it may make sense for us to take a dependency on eventually, though. No need to refactor now - let me have a closer look and then we'll see what makes sense 😄 This is exciting stuff!

You may also be interested in this thread where we discussed tag stuff. Obviously not enormously relevant since these are all `get` methods, but definitely curious to hear your thoughts on how you envision programmatically setting / managing tags.

https://community.rstudio.com/t/updating-tags-programatically/43199/5

tbradley1013:
I can't really take credit for the print formatting. I just got the output of the API into the same format that `fs::dir_ls` returns so that I could use their `dir_tree` implementation. So all credit to them! 

That makes sense about `stingr`, I knew there was an issue with either `stringr` or `stringi` but couldn't remember which. Now that I am thinking about it, I think we may be able to just use `dirname` since the strings I am creating are similar to filepaths... My only reservation with that is I am not sure if that method will be robust to different OS. 

That thread is interesting. I added a comment about potential confusion with more complex tag structures but I do like that idea on how to apply tags. I haven't thought too much about programmatically applying the tags but that method does seem pretty useful

tbradley1013:
Ok, so I figured out a pretty easy work around to avoid adding `stringr` as a dependency so I went ahead and implemented it so you can tackle the issue of adding `stringr` or `stringi` as a dependency later.

R/tags.R
```diff
@@ -0,0 +1,125 @@
...
+  
+  tag_tree(res, top_tag = top_tag)
+}
+
+tag_tree <- function(tags, top_tag = "tags"){
+  parsed_tags <- parse_tags(tags, top_tag = top_tag)
+  
+  tag_split <- split(parsed_tags, dirname(parsed_tags))
+  
+  ch <- fs:::box_chars()
```
colearendt: Unfortunately using unexported functions is a no-no, so we'll probably have to see if it's easy enough to pull these little utilities across into our package. It won't be the first utilities used from another package 😅 Although this reminds me that I should probably put attribution somewhere... 😬

R/tags.R
```diff
@@ -0,0 +1,125 @@
...
+  
+  print_leaf <- function(x, indent) {
+    leafs <- tag_split[[x]]
+    for (i in seq_along(leafs)) {
+      if (i == length(leafs)) {
+        cat(indent, fs:::pc(ch$l, ch$h, ch$h, " "), basename(leafs[[i]]), "\n", sep = "")
```
colearendt: Similarly for `fs:::pc`

R/tags.R
```diff
@@ -0,0 +1,125 @@
...
+        print_leaf(leafs[[i]], paste0(indent, "    "))
+      } else {
+        cat(indent, fs:::pc(ch$j, ch$h, ch$h, " "), basename(leafs[[i]]), "\n", sep = "")
```
colearendt: Again for `fs:::pc`

R/tags.R
```diff
@@ -0,0 +1,125 @@
...
+#' @param top_tag The value to be printed at the top of the 
+#' tag tree. This value can not have any "/" characters in it. 
+#' 
+#' @export
+get_tag_tree <- function(src, top_tag = "RStudio Connect Tags"){
+  validate_R6_class("Connect", src)
+  
+  res <- src$get_tag_tree()
+  
+  tag_tree(res, top_tag = top_tag)
```
colearendt: So this currently returns `NULL`, which seems a bit hard to "use." What are your thoughts on either:
- inventing an S3 class with a `print` method that gives you this pretty "tree" output
- making this more of a `print` or `show` method name, since you can't actually _use_ it for anything but illustration

If you look carefully, this is the pattern followed in `fs`. I'm also happy to accept this PR as-is and work on that change in a follow-up, if that's preferable, though! It's a bit of rearchitecting to be done 😄 This is a really cool picture of the nested tag list!

```
# fs example - sorry. Lazy and not a reprex
> hm <- fs::dir_tree("1558583138/")
1558583138/
> hm
1558583138/get_files
1558583138/get_files/rstudio-connect.gcfg
1558583138/get_files/rstudio-connect.gcfg/Failed
1558583138/get_files/rstudio-connect.gcfg/Failed/i-06fa8f5a99359338f
1558583138/get_files/rstudio-connect.gcfg/Failed/i-070464f64bc2fe5b2
1558583138/get_files/rstudio-connect.gcfg/Failed/i-08bffa9cb088d3093
1558583138/get_files/rstudio-connect.gcfg/Failed/i-09cf956e63c58753a
1558583138/get_files/rstudio-connect.gcfg/Failed/i-09d05131561dc044d
1558583138/get_files/rstudio-connect.gcfg/Failed/i-0b6a4a8df217ecb3d
1558583138/get_files/rstudio-connect.gcfg/Failed/i-0bbdd9a4ef5108a20
1558583138/get_files/rstudio-connect.gcfg/Failed/i-0c6933ba1d3d27a31
1558583138/get_files/rstudio-connect.gcfg/Failed/i-0c9dbdcc13dd5910b
1558583138/get_files/rstudio-connect.gcfg/Failed/i-0d3b4c10e58a50c55
1558583138/get_files/rstudio-connect.gcfg/Failed/i-0d3e90864c93d7992
1558583138/get_files/rstudio-connect.gcfg/Failed/i-0d4aa44abb1d5b83b
1558583138/get_files/rstudio-connect.gcfg/Failed/i-0d5160ac485a3fa60
1558583138/get_files/rstudio-connect.gcfg/Failed/i-0d616c0dc35382389
1558583138/get_files/rstudio-connect.gcfg/Success
1558583138/get_files/rstudio-connect.gcfg/Success/i-08497d121cbdd253a
1558583138/get_files/rstudio-connect.gcfg/Success/i-0a2030206cd446d46
> class(hm)
[1] "fs_path"   "character"

# connectapi
> obj <- get_tag_tree(client)
RStudio Connect Tags
├── Technology
│   ├── Deep Learning
│   │   ├── keras
│   │   ├── tensorflow
│   │   └── Pre-trained Models
│   ├── Output Types
│   │   ├── Excel
│   │   ├── PDF
│   │   └── PowerPoint
│   ├── Email
│   ├── BackEnds
│   │   ├── Databases
│   │   └── Spark
│   └── Dashboards
│       ├── flexdashboard
│       ├── shinydashboard
│       ├── D3
│       └── Maps
├── Projects and Presentations
│   ├── Calendars
│   ├── QuickStart
│   │   └── Stock Portfolio
│   ├── Production Webinar
│   │   ├── Production
│   │   └── Prototype
│   ├── Pins Demo: Bikes
│   │   ├── Pins
│   │   └── Other Assets
│   └── Pro Admin Training
└── Solutions
    ├── Access to Care
    ├── Python
    ├── Immunotherapy
    ├── Customer Churn
    └── Model Management
> obj
NULL
```

R/tags.R
```diff
@@ -0,0 +1,125 @@
...
+  tag_tbl <- parse_tags_tbl(res)
+  
+  return(tag_tbl)
+}
+
+#' Print a visualization of the tag tree
+#' 
+#' @param src The source object
+#' @param top_tag The value to be printed at the top of the 
+#' tag tree. This value can not have any "/" characters in it.
```
colearendt: Do you know off-hand if Connect can have slashes in tag names? I can go check if not 😅 Do tag names w/ slashes cause trouble in this parsing approach?

R/tags.R
colearendt: Related:
```
> fs:::print.fs_path
function (x, ..., max = getOption("max.print")) 
{
    x <- x[seq_len(min(length(x), max))]
    if (length(x) == 0) {
        print(character(0))
    }
    else {
        cat(multicol(colourise_fs_path(x, ...)), sep = "")
    }
    invisible(x)
}
<bytecode: 0x7f7fe143cdc8>
<environment: namespace:fs>
```

R/tags.R
tbradley1013: It looks like you can include slashes in your tags and it results in them being excluded from the tag tree... We can use a different separator rather than a `/` but it will mean that `dirname` and `basename` will not be able to be used to parse the tags

R/tags.R
tbradley1013: I removed dependencies with https://github.com/rstudio/connectapi/pull/37/commits/d7e7446eb7dd4ce1eb73ed19e649f29365b49725 

Curiously, is the "no unexported functions" due to the fact that they may change without any notice? I had never heard it before but it makes sense now that you say that.

R/tags.R
tbradley1013: removed with d7e7446

R/tags.R
tbradley1013: removed with d7e7446

R/tags.R
tbradley1013: What if we simply add `invisible(parse_tags_tbl(res))` to the end of the function which would return:

```
client <- connectapi::connect()
 
tmp <- get_tag_tree(client)
#> RStudio Connect Tags
#> +-- BLS
#> |   \-- SRA
#> \-- OOW
#>     \-- H&H
 
tmp
#> # A tibble: 4 x 6
#>      id name  created_time         updated_time         version parent_id
#>   <int> <chr> <chr>                <chr>                  <int>     <int>
#> 1     1 BLS   2019-10-24T18:48:01Z 2019-10-24T18:48:01Z       0        NA
#> 2     4 SRA   2019-10-24T18:48:21Z 2019-10-24T18:48:21Z       0         1
#> 3     2 OOW   2019-10-24T18:48:12Z 2019-10-24T18:48:12Z       0        NA
#> 4     3 H&H   2019-10-24T18:48:16Z 2019-10-24T18:48:16Z       0         2
```

tbradley1013: Or we could return the unparsed list too with `invisible(res)` if you think people would want to see it in the nested list format

colearendt:
Merging so we can start CI again... will probably need a few follow-on PRs to iron out the kinks

PR #95, opened 2021-03-04
Continuous Integration Testing migration to GitHub Actions

ian-flores:
This PR incorporates the migration of the Integration Testing of the package with RStudio Connect from using Jenkins to using GitHub Actions. It deletes the Jenkinsfile as part of this process and creates an integration-test workflow.


colearendt (approved):
Looks awesome!! Apologies for the flailing there at the end - I was hoping we could run these on our self-hosted runners, but we currently don't allow self-hosted runners for public repos (security).

PR #101, opened 2021-03-17
Flexible http requests

colearendt:
Trying to make extending the HTTP client easier... (i.e. for special cookies, headers, etc.)


colearendt:
We need to trace `httr_additions` through the `PUT`, `HEAD`, `POST` endpoints

colearendt:
This is a little scary, since we're changing stuff that is very low level in the package (`GET`, `POST` requests, etc.). In any case, it seems stable! Here goes nothing!

PR #103, opened 2021-04-12
Bump rsc to 1.8.8

colearendt:


colearendt:
Currently holding for a bug in 1.8.8 that makes our integration tests fail

Issue #94, opened 2021-03-04
Update endpoints available to RStudio Connect 1.8.6


ian-flores:
RStudio Connect 1.8.6 offers a new set of endpoints that are not all documented and/or included in the current version of the `connectapi` package. We can track the new endpoints and their documentation from here: https://docs.rstudio.com/connect/1.8.6/api/#overview

colearendt:
Some ideas:
- [Content APIs / Env Vars](https://docs.rstudio.com/connect/api/#tag--Content)
- [Content ACLs - we may need to change our API](https://docs.rstudio.com/connect/api/#tag--Content-Permissions)
- [Tag stuff](https://docs.rstudio.com/connect/api/#tag--Tags)
- [Task endpoint](https://docs.rstudio.com/connect/api/#tag--Tasks)
- [Vanity URLs](#113)

rstub:
It seems as manipulating the `run_as` user also uses the old API:
https://github.com/rstudio/connectapi/blob/e88deb55b47414467d2720c4c625cbbea6194732/R/content.R#L52-L64

PR #117, opened 2021-04-27
add permissions

colearendt:
These should either supersede or replace the backend for the `acl_` helpers.

Since the API now formally calls them `permissions`, perhaps we should adopt that standard...

Related to #94


R/content.R
```diff
@@ -543,3 +570,146 @@ delete_bundle <- function(connect, bundle_id) {
...
+  }
+}
+
+
+#' @rdname permissions
+#' @export
+get_content_permissions <- function(content) {
+  validate_R6_class(content, "Content")
+  res <- content$permissions()
+  connectapi::parse_connectapi_typed(res, !!!connectapi_ptypes$permissions)
```
edavidaja, line 711: Does this need namespacing this way?

colearendt: Oh no 😄 Woops! Good catch 😄

PR #119, opened 2021-05-04
Fix for use of `path` in bundle_dir

### LINKED ISSUE #118: Error using bundle_dir with a path in latest release
See comment on https://github.com/rstudio/connectapi/commit/7011ab2c147813e4441e40f9134c55dd16aaecae?branch=7011ab2c147813e4441e40f9134c55dd16aaecae&diff=split#r50359881

Currently seeing errors in our deployment pipeline when people upgrade to latest connectapi versions.

### LINKED ISSUE #121: fs_dir error
I'm getting an error when trying to bundle a subdirectory which originates from the `fs` package:

```
  rsconnect::writeManifest("docs",
                           appFiles = list.files("docs", recursive = TRUE))
[ENOENT] Failed to search directory 'docs': no such file or directory 
```
The source of this error is this [setwd() code](https://github.com/rstudio/connectapi/blob/b39c14de7029b6bb1dc9c61bb7aaca0782da0a83/R/deploy.R#L119-L121) which I think is unnecessary, is there any reason to do the working directory dance rather than passing path directly to check bundle contents?

slodge:
Fix for checking bundle after `setwd(path)` has been called

Close #118 
Close #121


slodge:
Linked to #118

R/deploy.R
```diff
@@ -120,8 +120,8 @@ bundle_dir <- function(path = ".", filename = fs::file_temp(pattern = "bundle",
   setwd(path)
   on.exit(expr = setwd(before_wd), add = TRUE)
 
-  message(glue::glue("Bundling directory {path}"))
-  check_bundle_contents(path)
+  message(glue::glue("Bundling current directory ({path})"))
```
colearendt, line 123: Perhaps we should move this up above `setwd` (and leave the wording as "Bundling directory"). I think "Bundling current directory" would be confusing to a user receiving the message if they are not bundling their current directory

colearendt (commented):
Thanks for the PR! Do you have any interest in adding a test to protect some of these behaviors? If not, I'm happy to add!

slodge:
Interest yes... time today (and the next few days) a bit limited though (as
in "I'm run off my feet")

colearendt:
No rush at all!! Thanks @slodge ! I'll keep this open for you - if we get to the point of the CRAN release (held up for now behind some Connect changes), then I will probably go ahead and finish it off 😄

colearendt:
@slodge I added a couple of tests and flipped the wording slightly, since we're nearing CRAN release! We would definitely love to have your contribution in any other area you have interest!! I would also be happy to chat through things if that would be helpful, or if you have ideas!

slodge:
Thanks.

Sorry for not getting the tests done. Work and life getting in the way.

Please don't confuse this with any man of enthusiasm!

Stuart

colearendt:
Definitely not at all!! 😄 And no need to be sorry! You did all of the hard work for me - identified the issue, tracked it down, and fixed it!

PR #130, opened 2021-08-06
Content permission

colearendt:
@kmasiello FYI - these helpers are experimental but I think they may help our friends!

Does this look like decent names / programming interface?


kmasiello:
Names / interface seems fine. Having a hard time testing it with much success.

Test on:
 ```
board_register("rsconnect", server = "https://colorado.rstudio.com/rsc")
pin_get("katie/all-content-perms", board = "rsconnect")
```

I think in `content_list_guid_has_access` the double bar `||` should be single `|` for correct filtering, otherwise I get zero results for my guid.
 ```
filtered <- content_list %>% filter(access_type == "all" || 
    access_type == "logged_in" || owner_guid == {
    {
      guid
    }
  } || {
    {
      guid
    }
  } %in% permission$principal_guid)
```

kmasiello:
I don't think the filter for guid_has_access should include `access_type == "logged_in"`. 
https://github.com/rstudio/connectapi/blob/0814dad75ea21014ad14f6c7ec5dd314fbf27531/R/get.R#L357

colearendt:
Thanks for testing! Sorry for not catching that 🙈 

The idea is that if the Access Control is set to "All Users" or "All Logged in Users" then this user will have access, regardless of permissions

kmasiello:
🎉

PR #168, opened 2022-09-20
Fix ssl verify off

### LINKED ISSUE #161: Verify invalid SSL cert handling
We have a report that the SSL cert check docs we created do not actually work 🙈 

We should verify whether or not that is the case, and fix it if it does not!

colearendt:
Close #161 

Ultimately hard to test in CI because we do not currently have an "SSL Connect with invalid cert" (although that is something that we could potentially add to CI at some point).


colearendt:
Closing in favor of #169

PR #173, opened 2022-10-01
Ci tweaks

colearendt:
Try to make CI easier to detect issues with!

TODO: set up slack webhook


nealrichardson:
Looks like the same as #178, we can try again in #240.

PR #174, opened 2022-10-01
fix typo in message glue string

MichaelChirico:
Better to construct the glue input separately for a complicated expression with lots of delimiters to avoid this issue.

Fixes https://github.com/r-lib/lintr/issues/1459


colearendt (approved):
This is great! And great catch - thank you!

PR #176, opened 2022-11-01
Fix for dev purrr

hadley:
The recycling error now comes from vctrs so has a different class. I made the minimum change here, but testing the specific error class is a little brittle, so I'd suggest changing to snapshot tests at some point in the future.


hadley:
FYI purrr is now scheduled for release to CRAN on Dec 19

PR #178, opened 2022-11-22
Fix a few

colearendt:


nealrichardson:
This isn't mergeable, and the image names have evolved from this. I made #240 to capture the idea, we can revisit.

PR #182, opened 2023-01-26
Standardize Posit name

jonthegeek:
I scraped CRAN to find companies that fund R packages, and noticed that Posit is credited 6 different ways (9 if you include RStudio/RStudio PBC/RStudio, PBC). I'm submitting PRs with the official "Posit Software, PBC" name on the few Posit/PBC varieties to nip it in the bud.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -10,7 +10,7 @@ Authors@R:
              family = "Arendt",
              role = c("aut", "cre"),
              email = "cole@rstudio.com"),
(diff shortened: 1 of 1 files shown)


colearendt:
Thanks for this!! 🎉 https://github.com/r-lib/usethis/issues/1683 We are doing a bit of debate around whether we want to standardize on `Posit, PBC` or `Posit Software, PBC`. But I can take care of that! 😄

PR #189, opened 2023-03-22
use the public API for bundle download and deletion

aronatkins:
Tested with code like the following:

```r
library(connectapi)
client <- connect()
content_guid <- "ABCD-ABCD-ABCD-ABCD-ABCD"
content <- connectapi::content_item(client, content_guid)
bundles <- connectapi::get_bundles(content)
bundle_id <- tail(bundles, n=1)$id
connectapi::download_bundle(content, bundle_id = bundle_id)
connectapi::delete_bundle(content, bundle_id = bundle_id)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # connectapi (development version)
 
+- BREAKING: The `download_bundle()` and `bundle_delete()` functions require a
+  `content` item as well as the target `bundle_id`. Both use the public
+  API for bundle management, which has been available since Connect 1.8.6.
+
 # connectapi 0.1.3.1
 
 - Fix generated documentation HTML for CRAN submission
--- a/R/connect.R
+++ b/R/connect.R
@@ -393,16 +393,14 @@ Connect <- R6::R6Class(
       )
     },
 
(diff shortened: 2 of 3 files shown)


christierney (approved):

colearendt (changes requested):
I'm not sure we actually need this? 

The reason for the `deprecate_soft` is that the `Connect$download_bundle()` and `Connect$delete_bundle()` are no longer used / recommended.

Instead, it lives on the `Content` object 😅 Curious on y'all's thoughts for this / whether that is confusing or we could make this clearer. I am inclined to just `deprecate_hard()` these functions you changed or remove them.

https://github.com/rstudio/connectapi/blob/c45117db71c688f23c827ea2468684ee59cc3402/R/content.R#L37-L45

aronatkins:
I'm trying to prove that we can eliminate some of the `../v1/experimental` APIs and having them used by `connectapi` blocks that work.

I'm happy to have these functions disappear entirely. Is that work you already have planned?

aronatkins:
Closing this PR in favor of one that removes these functions entirely.

PR #190, opened 2023-03-28
remove experimental bundle use

aronatkins:
Replaces https://github.com/rstudio/connectapi/pull/189.


man/PositConnect.Rd
colearendt, line 1: Why did this one get deleted? 😮

colearendt (approved):
LGTM!

man/PositConnect.Rd
aronatkins: These are the functions that we deleted. Most of the file is still there; GitHub is just choosing to hide the generated doc files from you.

colearendt: 🤦 Wow, that's a wild miss. Sorry about that! Not sure how/why I interpreted that as the file getting deleted

Issue #202, opened 2023-08-01
Bug: when there is more than one app with the same name the code breaks


cormach:
When running connectapi::deploy(), the code fails when there is more than one app with the same name in the Posit Connect instance.  

https://github.com/rstudio/connectapi/blob/main/R/content.R#L536

The error is 
```
Error in eval(parse(text = text, keep.source = FALSE), envir) : 
  object 'to_content' not found
Calls: <Anonymous> -> content_ensure
```
The code currently references is as follows:

```
if (length(content) > 1) {
      stop(glue::glue(
        "Found {length(to_content)} content items ",
        "matching {name} on {connect$server}",
        ", content must have a unique name."
      ))
```
The problem occurs because to_content isn't defined

PR #203, opened 2023-08-01
Fix error message in content_ensure()

cormach:
#202 solving this issue


nealrichardson (approved):
Thank you for this fix. It looks like there is no test covering this case yet, but we can add one separately.

PR #207, opened 2024-01-04
re-document and fix warnings

aronatkins:
CRAN is currently reporting these warnings and they have caused PR tests to fail.
https://cran.r-project.org/web/checks/check_results_connectapi.html


colearendt (approved):
You are a champion! Thanks for tracking this down!

PR #212, opened 2024-02-14
Fix typo

olivroy:
just to make sure it doesn't fall through the cracks.


hadley (approved):

hadley:
@aronatkins yeah, I wish there was some way to do that more easily. But in this case, I don't think we need to worry since it will only be triggered in the case where dbplyr is not installed, which I don't think any of our builds cover. But we could add a mocked test if you wanted.

PR #223, opened 2024-04-05
Introduce httptest for API mocking; update to testthat 3e

nealrichardson:
This PR adds https://enpiar.com/r/httptest/ for testing the API client with mock/recorded API responses. This allows contributors to run tests for the R code without having to run a local Connect server, which is not an option for most. I'd like to get some good overall coverage of the package's behaviors before doing any refactoring, just so we have some confidence that we aren't breaking things.

Before this PR, the line coverage of the non-integration tests was 15%. I've added some basic tests for Content, Permissions, and Users, pulling in the fixtures from posit-sdk-py. This gets coverage up to 24%. Followup PRs can extend this approach to get more coverage.

I avoided making changes in the package itself, though I did fix a bug in an error message builder, one broken API doc link, and I added `Config/testthat/edition: 3` to the DESCRIPTION, both of which triggered some autoformatting. 

I also removed the `mockery` dependency, as it was only being used in one test that could be rewritten to use a mock request.


nealrichardson:
Integration test failures say the license has expired: https://github.com/rstudio/connectapi/actions/runs/8571214328/job/23490961358?pr=223#step:7:28

@colearendt @aronatkins can we fix this?

nealrichardson:
Integration tests are running now that this is rebased on main, but now they're failing, I'm guessing it has to do with testthat 3e. I'll fix.

PR #228, opened 2024-04-11
Use license file for integration tests

### LINKED ISSUE #227: Use license file for integration tests in CI
build_test_env() seems to be set up to use either a license key or a license file: https://github.com/rstudio/connectapi/blob/fbcfbb189d292923eed5dd3f5a9842ba912c2f89/R/utils-ci.R#L181

If the first argument looks like a path to a license file, it goes down one path, otherwise it assumes it is a key: https://github.com/rstudio/connectapi/blob/fbcfbb189d292923eed5dd3f5a9842ba912c2f89/R/utils-ci.R#L32-L50

It may be the case that if `RSC_LICENSE` is updated in GH secrets to be a file path, and a license file contents is included in another secret, and we add a step prior to `build_test_env()` that writes the contents of that second env var to `$RSC_LICENSE`, it will just work.

nealrichardson:
Fixes #227

It took a bit of trial and error to get the secret passed in correctly, and among the things I upgraded along the way:

* Switch `docker-compose` to `docker compose`
* Update workflows to use ubuntu-latest
* Remove `processx` dependency in utils-ci.R. and other simplification in that script

Followups:

* #230 
* #231


nealrichardson:
This works locally for me: I can run the same [docker-compose command](https://github.com/rstudio/connectapi/actions/runs/8652136522/job/23724355168?pr=228#step:8:23) to start Connect, having set the RSC_VERSION and RSC_LICENSE_FILE env vars appropriately, and when I hit the ping endpoint I get a 200 response and not 402. For some reason the license file isn't being picked up correctly in CI, and I'm not sure how to get more insight into what's happening in the container. Need to step away and poke at this some more later. @aronatkins lmk if you notice any gotchas I'm missing.

.github/workflows/integration-tests.yaml
```diff
@@ -15,6 +15,7 @@ jobs:
       RSC_VERSION: 2022.09.0
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       RSC_LICENSE: ${{ secrets.RSC_LICENSE }}
```
nealrichardson, line 17: I updated this secret in GH so that it is now a file path where a license file can be expected, not a license key.

R/utils-ci.R
```diff
@@ -34,7 +34,7 @@ determine_license_env <- function(license) {
     cat_line("determine_license: looks like a license file")
     return(list(
       type = "file",
-      cmd_params = c("-v", paste0(license, ":/etc/rstudio-connect/license.lic")),
```
nealrichardson, line 37: This is the only change in this file, the rest is reformatting triggered on save.

.github/workflows/integration-tests.yaml
aronatkins: In other contexts (Connect testing), we use `CONNECT_LICENSE` for a key and `CONNECT_LICENSE_FILE` for the file. That lets us understand if we should `license-manager activate` or copy the file to the landing place.

The Connect Docker images uses `RSC_LICENSE` for a key and `RSC_LICENSE_FILE_PATH` for file. That's probably what inspired this workflow. https://github.com/rstudio/rstudio-docker-products/blob/d075a3c0a7bd8efc6b679ae33eceaa905234c30b/connect/startup.sh#L32-L40

.github/workflows/pkgdown.yaml
```diff
@@ -42,6 +43,9 @@ jobs:
           remotes::install_github("r-lib/lifecycle")
         shell: Rscript {0}
 
+      - name: Set up license file
+        run: echo ${CONNECT_LICENSE_FILE} > $RSC_LICENSE
```
aronatkins, line 47: Permissions are unlikely to matter in this context, but in case you want to see what we recommend: https://docs.posit.co/connect/admin/licensing/#license-file

R/utils-ci.R
```diff
@@ -60,8 +62,7 @@ compose_start <- function(connect_license = Sys.getenv("RSC_LICENSE"), rsc_versi
   compose_path <- find_compose()
 
   license_details <- determine_license_env(connect_license)
-  compose_file <- switch(
-    license_details$type,
+  compose_file <- switch(license_details$type,
     "file" = "ci/test-connect-lic.yml",
```
aronatkins, line 66: Future: We should squash these compose files into one (or remove them entirely) and simply provide the correct bind mounts and environment variables to the container as it starts.

aronatkins (approved):
Looks fine based on my understanding, but I don't have any history with this part of the CI configuration.

Later: The compose files may want to shift to using an OS other than Ubuntu Bionic.

.github/workflows/integration-tests.yaml
nealrichardson: Yeah I was going to rename a bunch of things once I got it working, I agree it's not ideal as it is.

.github/workflows/pkgdown.yaml
nealrichardson: The file permissions look to be the same as what I have locally, which works. IDK about the user.

R/utils-ci.R
nealrichardson: This is no longer true, I simplified a bit more when I switched from `docker-compose` to `docker`

.github/workflows/pkgdown.yaml
nealrichardson: After a bunch of attempts, it turns out I needed `echo "$CONNECT_LICENSE_FILE"` with quotes in order to correctly handle the newlines in the license file-in-env-var. This is working now.

nealrichardson:
Remaining CI failures are unrelated. I'll explore them separately.

PR #236, opened 2024-04-16
Run styler on the whole codebase

nealrichardson:
My editor formats on save, and that means when I touch one line in a file, it makes a big diff. So let's just run `styler` on the whole package and get it over with. No sense in arguing with the robots about what correct code style is.

### DIFF
--- a/R/audits.R
+++ b/R/audits.R
@@ -39,7 +39,6 @@ cache_apps <- function(connect) {
 #' @family audit functions
 #' @export
 audit_vanity_urls <- function(apps, server_url, vanity = NULL) {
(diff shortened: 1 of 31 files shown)


jonkeane:
I'm fine with this, but to confirm: does your editor format in a style that styler agrees wtih?

nealrichardson:
> I'm fine with this, but to confirm: does your editor format in a style that styler agrees wtih?

Yes, VSCode uses `styler`: https://github.com/REditorSupport/languageserver?tab=readme-ov-file#customizing-formatting-style

Issue #219, opened 2024-03-29
add `vanity_url` to `/v1/content` output 


kmasiello:
`/v1/content` provides `dashboard_url` and `content_url` but does not provide the vanity URL, if set. 

The old `get_content_old()` function would provide a TRUE/FALSE value for `vanity_url` and reported the url as the `content_url`.

The ergonomics of doing multiple API calls to get content info, get list of vanity url, and then joining is awkward.

jonkeane:
I need to dig further, but I suspect this might actually be a bug in the API of connect itself. From the API docs:

```
content_url: string The URL associated with this content. Computed from the associated vanity URL or GUID for this content.
```

Which to me reads as: the content_url will be the vanity if a vanity has been set or the GUID version if not, but when I tried it myself just now on content I know has a vanity I got back only the GUID. Why else would the docs mention vanity URL at all? But I'll need to dig into that and see if I can figure out what's wrong there.

This is a bit annoying, but you could also use `connectapi::get_vanity_url()` (though you have to construct a content item from each guid you have with `connectapi::content_item()`, and also concatenate the base URL onto the vanity suffix you get back)

jonkeane:
Should have reported back here earlier — it _turns out_ that the docs were wrong. I've updated them, but vanities are not available there. They could (and probably should) be added to the API, but they are not there today.

So we'll have to do something like this in the meantime :( 

> This is a bit annoying, but you could also use connectapi::get_vanity_url() (though you have to construct a content item from each guid you have with connectapi::content_item(), and also concatenate the base URL onto the vanity suffix you get back)

PR #245, opened 2024-04-19
Remove long-deprecated functionality

### LINKED ISSUE #242: Remove long-deprecated functionality
Anything that has been flagged as deprecated for years should be deleted.

nealrichardson:
Fixes #242


R/get.R
```diff
@@ -376,152 +376,6 @@ content_list_guid_has_access <- function(content_list, guid) {
...
-#' client <- connect()
-#'
-#' get_content_old(client, limit = 20)
-#' }
-#'
-#' @export
-get_content_old <- function(src, filter = NULL, limit = 25, page_size = 25) {
-  validate_R6_class(src, "Connect")
-
-  lifecycle::deprecate_warn("0.1.0.9023", "get_content_old()", "get_content()", details = "The filter argument is deprecated and the structure of the response has changed with the public API")
```
jonkeane, line 508: This is likely more a comment about the deprecation warning, but this makes it sound like _only_ the `filter` argument is deprecated? 

But given the function name, that must be wrong / I'm fine removing this.

I don't think this is worth keeping it for, but one note that #219 mentions this being a place one can acquire vanities without looping through

jonkeane (approved):
🚮

R/get.R
nealrichardson: I see the concern but I think this still should be removed.

* The deprecation warning says that `get_content_old()` is deprecated and `get_content()` is the new function you should call.
* The API doc link it has is broken. The API it's hitting is the "old" undocumented one.
* It is not tested.

We should solve the vanity URL issue in the API better.

PR #246, opened 2024-04-19
Move dplyr to Suggests

nealrichardson:
It's only truly needed for `tbl_connect`. 

We should reduce our dependency footprint so that the package is cheaper to install for cases where someone just needs to do a simple task in the API.


R/get.R
```diff
@@ -367,13 +371,10 @@ content_list_by_tag <- function(src, tag) {
...
-    access_type == "all" |
-      access_type == "logged_in" |
-      owner_guid == {{ guid }} |
-      purrr::map_lgl(permission, ~ {{ guid }} %in% .x$principal_guid)
-  )
-  return(filtered)
+  row_filter <- content_list$access_type %in% c("all", "logged_in") |
+    content_list$owner_guid == guid |
+    purrr::map_lgl(content_list$permission, ~ guid %in% .x$principal_guid)
+  content_list[row_filter, ]
```
jonkeane, line 377:
```suggestion
  rows_keep <- content_list$access_type %in% c("all", "logged_in") |
    content_list$owner_guid == guid |
    purrr::map_lgl(content_list$permission, ~ guid %in% .x$principal_guid)
  content_list[rows_keep, ]
```

This is super minor, but `row_filter` didn't totally connect with me at first what's going on here (and has the awkward semantics of where `filter == TRUE`, that means keep — not unique here, and totally what `dplyr:filter` does) . Other possibilities: `row_index` `rows_to_keep`

Also fine as is, it's not a huge problem

R/lazy.R
```diff
@@ -58,8 +58,7 @@ check_deprecated_names <- function(.name, deprecated_names) {
   return(.name)
 }
 
-#' @importFrom dplyr collect
-#' @export
+# This will be registered in .onLoad is dplyr is available
```
jonkeane, line 61:
```suggestion
# This will be registered in .onLoad if dplyr is available
```

I think this is what you intended?

R/lazy.R
```diff
@@ -118,7 +117,7 @@ print.tbl_connect <- function(x, ..., n = NULL) {
 
 #' @export
 as.data.frame.tbl_connect <- function(x, row.names = NULL, optional = NULL, ..., n = Inf) {
-  as.data.frame(collect(x, n = n))
+  as.data.frame(dplyr::collect(x, n = n))
 }
```
jonkeane, line 121: Might be worth an explicit comment that we don't need to condition on this because the only way to have `tbl_connect` available at all is if dplyr is already available.

R/tags.R
```diff
@@ -435,7 +435,7 @@ tag_tree <- function(.x) {
 
 parse_tags_tbl <- function(x) {
   parsed_tags <- purrr::map_dfr(x, ~ {
-    out <- dplyr::tibble(
+    out <- tibble::tibble(
```
jonkeane, line 438: We can later, but is it worth doing the hacky add the class on without also requiring `tibble`?

jonkeane (approved):
Yay! Glad to see this. `Magrittr` is much smaller, but would it make sense to 🚮 the pipe import here too?

nealrichardson:
> Yay! Glad to see this. `Magrittr` is much smaller, but would it make sense to 🚮 the pipe import here too?

It's used a few other places. I can remove it separately.

R/lazy.R
```diff
@@ -118,7 +117,7 @@ print.tbl_connect <- function(x, ..., n = NULL) {
 
 #' @export
 as.data.frame.tbl_connect <- function(x, row.names = NULL, optional = NULL, ..., n = Inf) {
-  as.data.frame(collect(x, n = n))
+  as.data.frame(dplyr::collect(x, n = n))
```
nealrichardson, line 120:
```suggestion
  # We don't need to check if dplyr is available here
  # because you won't have a tbl_connect without first
  # checking for dplyr.
  as.data.frame(dplyr::collect(x, n = n))
```

R/tags.R
nealrichardson: I'm not sure, tibble is used heavily throughout, they're the main thing that the API wrappers return. Can make an issue to evaluate.

nealrichardson:
> > Yay! Glad to see this. `Magrittr` is much smaller, but would it make sense to 🚮 the pipe import here too?
> 
> It's used a few other places. I can remove it separately.

FWIW other packages we depend on also import magrittr, so even if we remove our direct dependency on it, it won't have an effect on the total number of packages required to be installed (at least today).

PR #247, opened 2024-04-19
Remove some unused functions

nealrichardson:
Looks like an interesting idea was brewing here, but nothing came of it, and these functions are not documented, exported, or tested. 

The idea of recording deployment configuration in a YAML file is a good one, and we've picked it up in the new publisher project. Whenever we want to revisit that idea here, we should follow the specification they're defining rather than invent another new one.


jonkeane (approved):
😱 at no tests

PR #261, opened 2024-05-16
Allow exact group matches

### LINKED ISSUE #216: Trying to create groups but search is returning two simular group names
We have encountered a problem while using the `groups_create_remote` method in the ConnectAPI R package to create groups. The issue arises when the method detects two groups with similar prefixes, where one group's name is the prefix of the other. Due to this, we are unable to create the groups using the provided API.

For example, consider the made-up group names `ABE` and `ABE-TEST`. When we create the group using the groups_create_remote method for `ABE` the method fails as `ABE` is also the prefix of the `ABE-TEST` group.

Previously, the same problem have been flagged in the `users_create_remote` method which have been solved in: https://github.com/rstudio/connectapi/issues/135

Thanks

jonkeane:
Adds an `exact` parameter to `groups_create_remote` to allow creating a specific group even if there are multiple matches. This follows a similar change in #157 

This as well as the matching `users_create_remote` could use some refactoring, though I'm not sure it's worth it in this PR.

Resolves #216


tests/testthat/test-remote.R
```diff
@@ -0,0 +1,34 @@
+with_mock_api({
+  con <- Connect$new(server = "https://connect.example", api_key = "fake")
+
+  test_that("groups_create_remote errors if prefix returns zero matches", {
+    skip("not implemented")
```
nealrichardson, line 5: Why not?

tests/testthat/test-remote.R
```diff
@@ -0,0 +1,34 @@
...
+      "Creating remote group"
+    )
+  })
+  test_that("groups_create_remote creates match if one is found", {
+    skip("not implemented")
+  })
+  test_that("groups_create_remote creates multiples if multiple are found", {
+    expect_error(
+      groups_create_remote(con, "Art"),
+      "The expected group\\(s\\) were not found"
```
nealrichardson, line 25: I don't understand how this matches the test description

anders-wiggers:
Hi,
What is the status of this PR? Is it stuck? Is there anything we can assist with to complete the feature? This is required for our enterprise solution.

tests/testthat/test-remote.R
```diff
@@ -0,0 +1,34 @@
...
+  })
+  test_that("groups_create_remote errors if expect is greater than one", {
+    skip("not implemented")
+  })
+  test_that("groups_create_remote works if match already exists", {
+    expect_message(
+      groups_create_remote(con, "Everyone"),
+      "Creating remote group"
+    )
+  })
```
toph-allen, line 18: Is this supposed to test the `check == FALSE` behavior?

tests/testthat/test-remote.R
```diff
@@ -0,0 +1,34 @@
...
+  test_that("groups_create_remote creates match if one is found", {
+    skip("not implemented")
+  })
+  test_that("groups_create_remote creates multiples if multiple are found", {
```
toph-allen, line 22: The test wording seems to suggest that it's intended to test the behavior when `expect > 1`. But there's also the skeleton of a test for an error if `expect > 1`.

R/remote.R
```diff
@@ -71,31 +71,39 @@ users_create_remote <- function(connect, prefix, expect = 1, check = TRUE, exact
...
+    }
     if (nrow(local_groups) > 0) {
-      message(glue::glue("At least one group with name prefix '{prefix}' already exists"))
+      if (!exact) {
+        message(glue::glue("At least one group with name prefix '{prefix}' already exists"))
+      } else {
+        message(glue::glue("A group with the name '{prefix}' already exists"))
+
+      }
       return(local_groups)
```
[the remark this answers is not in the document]
toph-allen: Outside the scope fo this PR.

toph-allen:
I made the following changes, and could use feedback on a few things:

**Testing**. Some tests were implemented using the `MockConnect` class. I was able to use some of the existing JSON files to get existing tests to work. There are tradeoffs to each workflow. Right now, upsides of specifying mocks in-line are (1) I find it easier to understand what information is returned by any given call, (2) the ability to have multiple different pieces of data returned from a given URL, (3) verifying the routes that were called. The main functional downside is that it lacks verification of the expected request body or query string, which I'd like to address. The more verbose tests are perhaps a downside too; there's definitely a tradeoff between that and having easily-discoverable context. I can leave it as-is or convert all the tests to a single type. (I would find this easier to go in the direction of `mock_dir` to in-line.)

**Return type**. The functions used to return a single user or group, which was not guaranteed to even be the created group (since search happens by prefix there's no guarantee that the returned group was gonna be the one created). I changed this to return the aggregated output of the group/user creation API calls. However, because these could potentially be multiple calls, this is currently a list of lists. This is also technically a breaking change. 🤔 

Three possible solutions I can think of:

1. Leave it as is. Address in a future PR where groups are actually given a class.
2. If we're making a breaking change, break it even more! Have the functions just return a data frame with each line representing a created user or group. That's at least nicer to look at and work with, and is probably what R users expect.
3. Cap the value of `expect` at 1. This would prevent the creation of multiple users or groups with a single function call and allow us to revert to the old return type. There was a not-implemented test that suggested this might have been someone's intent at some point, but the stated expectation of the test and the behavior of the code differed.

toph-allen:
@schloerke Thanks for your suggestions! Do you have any thoughts on my brain-dump in [this comment](https://github.com/rstudio/connectapi/pull/261#issuecomment-2439026624)?

R/remote.R
```diff
@@ -58,50 +58,61 @@ users_create_remote <- function(connect, prefix, expect = 1, check = TRUE, exact
...
 #'
 #' @export
-groups_create_remote <- function(connect, prefix, expect = 1, check = TRUE) {
+groups_create_remote <- function(connect, prefix, expect = 1, check = TRUE, exact = FALSE) {
   expect <- as.integer(expect)
-  if (check && expect > 1) {
-    stop(glue::glue("expect > 1 is not tested. Please set expect = 1, and specify a more narrow 'prefix'. You provided: expect={expect}"))
-  }
   if (check) {
     # TODO: limit = 1 due to a paging bug in Posit Connect
```
nealrichardson, line 79: Doesn't have to be done in this PR, but would be good to `git blame` this comment and look to see if this issue still/actually exists in Connect.

tests/testthat/test-remote.R
```diff
@@ -0,0 +1,244 @@
+with_mock_api({
```
nealrichardson, line 1: Are you even using this?

R/remote.R
toph-allen: Created an issue to track: https://github.com/rstudio/connectapi/issues/319

tests/testthat/test-remote.R
toph-allen: Yes, for a few tests — @schloerke and I re-ordered tests to make it clear what's using the mock API and what isn't.

toph-allen:
Reverted the return value after investigating `get_groups` with @schloerke. Updated incorrect docs for `get_groups` as a result. :)

toph-allen (commented):
Accidentally re-requested review from myself; getting it off my list.

R/remote.R
```diff
@@ -58,44 +58,55 @@ users_create_remote <- function(connect, prefix, expect = 1, check = TRUE, exact
     src = connect
   )
   message("Done creating remote users")
-  return(get_users(connect, prefix = prefix))
+  return(user_creation)
```
schloerke, line 61: Should this be reverted?

NEWS.md
```diff
@@ -7,6 +7,12 @@
 - New functions `set_thumbnail()`, `get_thumbnail()`, `delete_thumbnail()` and
   `has_thumbnail()` let you interact with content thumbnails, replacing older
   `*_image()` functions. (#294, #304)
+- `groups_create_remote()` now has an `exact` parameter. Setting `exact` causes
+  the function to consider only exact group name matches when searching for
+  remote groups and checking for existing local groups. (#216)
+- Previously, `groups_create_remote()` and `users_create_remote()` could return
+  a user or group other than the group(s) created. Now they always return the
+  details of the newly-created user(s) or group(s) in a list.
```
schloerke, line 15: `groups_create_remote()` was changed back and should be removed from the entry
```suggestion
- Previously, `users_create_remote()` could return a user or group other than the
  group(s) created. Now they always return the details of the newly-created user(s)
  or group(s) in a list.
```

If `users_create_remote()` is also changed, this bullet should be removed.
```suggestion
```

NEWS.md
```diff
@@ -7,6 +7,12 @@
 - New functions `set_thumbnail()`, `get_thumbnail()`, `delete_thumbnail()` and
   `has_thumbnail()` let you interact with content thumbnails, replacing older
   `*_image()` functions. (#294, #304)
+- `groups_create_remote()` now has an `exact` parameter. Setting `exact` causes
```
tdstein, line 10: Should this mention anything about the change to `users_create_remote` as well?

tdstein (approved):

R/get.R
```diff
@@ -59,7 +59,7 @@ get_users <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
 #' @param page_size the number of records to return per page (max 500)
 #' @param prefix Filters groups by prefix (group name).
 #' The filter is case insensitive.
-#' @param limit The max number of groups to return
+#' @param limit The max number of pages to return.
```
nealrichardson, line 62: Hmm, `limit` is typically about result count, not pages. If the paginated requester is not respecting `limit` correctly, we should fix that instead of documenting surprising behavior.

R/remote.R
toph-allen: Good catch, I only reverted the groups one.

NEWS.md
toph-allen: Accepting the suggestion locally  :)

NEWS.md
toph-allen: No — those changes were in a previous PR (from two years ago). The only change to `users_create_remote` was reverted.

R/get.R
toph-allen: Yeah, in the moment, @schloerke were experimenting with `get_groups()` and thought that the documentation was wrong. Digging deeper into the code, it just looks like there is a bug in the pagination… implementation? I need to dig into it, because it looks like it might have been around for a while. Tracked by #319, and I think perhaps high enough priority for me to look at it next week.

In the meantime I'll revert this; thanks!

Issue #221, opened 2024-04-05
Increase default page size to max


nealrichardson:
The users API, for example, is paginated, and the max page size is 500, but connectapi has a default value of 20. The user payload is not very large, so returning 500 records at once shouldn't be a problem, and it would definitely be cheaper than making 25 HTTP requests to collect them all.

In posit-sdk-py, we went with making the max the default.

PR #264, opened 2024-05-17
Use httr for URL encoding of query params

### LINKED ISSUE #263: Query params are not URL encoded
Using the `prefix` arguement for `get_groups()` works fine when there are no spaces, however with a space (and certain other symbols) will result in a malformed URL error (URL rejected: Malformed input to a URL function).

```r
client <- connect()
get_groups(client, prefix ="Test Group")
```

This applies beyond groups, see comments in the issue.

nealrichardson:
Fixes #263. I also hunted for everywhere else we pass in query parameters and stripped out the string interpolation, even if they only expected numbers and thus weren't needing URL encoding.

Along the way I also inserted some URL construction helpers that will have the side benefit of allowing us to more easily see where there are undocumented APIs being called, so that we can work to upgrade them here and in Connect. (Current score: 50 v1 calls, 1 v1/experimental, and 31 unversioned.)

I also changed the default page size to 500, the max, to save extra HTTP back and forth (#221).

~~I didn't catch *every* URL/query construction in this PR, but I got a lot of them, and I'll make a followup to get the rest when I have more time.~~ Actually I did, and I went through and added basic tests that made sure they were making the expected HTTP requests, at least for those that aren't covered already by integration tests. This has boosted test coverage to over [85% of lines](https://app.codecov.io/github/rstudio/connectapi/pull/264), IIUC.


R/connect.R
```diff
@@ -730,23 +712,13 @@ Connect <- R6::R6Class(
...
-          gsub("&+", "&", .)
+    groups_remote = function(prefix = NULL, limit = 500) {
+      path <- v1_url("groups", "remote")
+      query <- list(
+        limit = valid_page_size(limit),
+        prefix = prefix
       )
-
-      self$GET(path)
+      self$GET(path, query = purrr::discard(query, is.null))
```
jonkeane, line 721: This is fine as is since it's just a single function — but it does seem a little funny to have ever single query wrapped in this without some sort of helper (or we could override the `GET`, but that seems a little silly). 

But also, I thought `httr` did this stripping for us? But maybe some of [the setup and executing](https://github.com/rstudio/connectapi/pull/264/files#diff-5a4fc998698f639d47f38a272c1f3b10c3cd8316c72f0259037259dfbdd31369R141-R157)?

```
> query <- list(foo = NULL, bar = 1)
> httr::GET("https://jonkeane.com", query = query)
Response [https://jonkeane.com/?bar=1]
...
```

Anyway, not a huge must change, but it seems a little off

R/variant.R
```diff
@@ -38,12 +39,11 @@ Variant <- R6::R6Class(
...
     send_mail = function(to = c("me", "collaborators", "collaborators_viewers")) {
       warn_experimental("send_mail")
-      if (length(to) > 1) to <- "me"
-      url <- glue::glue("variants/{self$get_variant()$id}/sender")
+      url <- unversioned_url("variants", self$get_variant()$id, "sender")
       self$get_connect()$POST(
         path = url,
         body = list(
-          email = to
+          email = match.arg(to)
```
jonkeane, line 46: If we wanted to not allow partial matches (especially since the second is a substring of the last) we already have `rlang` elsewhere, we could use [`arg_match()`](https://rlang.r-lib.org/reference/arg_match.html)

R/connect.R
```diff
@@ -340,12 +340,6 @@ Connect <- R6::R6Class(
       tag_tree_parse_data(raw_tags)
     },
 
-    #' @description Get the tag tree.
-    get_tag_tree_old = function() {
-      warn_experimental("get_tag_tree")
-      self$GET("tag-tree")
-    },
```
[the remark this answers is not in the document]
nealrichardson: Yeah, just deleted. Not tested, not used in any other function, has the `_old` suffix.

R/connect.R
nealrichardson: I didn't even think to try, I'll strip these out and make sure tests pass.

R/variant.R
nealrichardson: `match.arg()` is probably fine here (and certainly better than how this was handled before) but yeah `arg_match()` is probably better.

R/connect.R
nealrichardson: You're right, it just worked 🎉

PR #268, opened 2024-05-23
Check for X-Deprecated-Endpoint header

nealrichardson:
`connectapi` shouldn't be calling deprecated endpoints--when we deprecate endpoints that it uses, we'll make sure to check the Connect server version and call the correct endpoints accordingly, so that it only calls deprecated endpoints on older Connect versions where they aren't deprecated. The only circumstance where `connectapi` would hit a deprecated endpoint would be if you're pointing at a newer version of Connect but an old version of `connectapi` that doesn't know about the new, non-deprecated endpoints. So, when receiving a `X-Deprecated-Endpoint` response header, the recommendation to the user is to upgrade `connectapi`.

### DIFF
--- a/R/connect.R
+++ b/R/connect.R
@@ -152,8 +152,8 @@ Connect <- R6::R6Class(
           self$httr_additions
         )
       )
-      check_debug(url, res)
-      return(res)
+      check_debug(res)
+      res
     },
 
     #' @description Perform an HTTP PUT request of the named API path. Returns an object parsed from the HTTP response.
@@ -182,7 +182,7 @@ Connect <- R6::R6Class(
         )
       )
       self$raise_error(res)
-      check_debug(req, res)
+      check_debug(res)
       httr::content(res, as = "parsed")
     },
 
@@ -203,8 +203,8 @@ Connect <- R6::R6Class(
           self$httr_additions
         )
       )
-      check_debug(req, res)
-      return(res)
+      check_debug(res)
+      res
     },
 
     #' @description Perform an HTTP DELETE request of the named API path. Returns the HTTP response object.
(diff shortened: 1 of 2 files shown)


R/connect.R
```diff
@@ -962,9 +962,21 @@ connect <- function(
...
+  # connectapi will make the right request based on the version of the server,
+  # but if you have an old version of the package, it won't know the new URL
+  # to request.
+  if ("X-Deprecated-Endpoint" %in% names(httr::headers(res))) {
+    msg <- paste(
+      res[["request"]][["url"]],
+      "is deprecated and will be removed in a future version of Connect.",
+      "Please upgrade `connectapi` in order to use the new APIs."
+    )
+    rlang::warn(msg, class = "deprecatedWarning")
```
aronatkins, line 977: This feels similar to the experimental warning, which happens at most once per session. Should this follow the same rules? See `warn_experimental` and `warn_once`.

tests/testthat/test-connect.R
```diff
@@ -39,6 +39,21 @@ test_that("Handling error responses", {
   expect_error(con$raise_error(resp), "Bad Request")
 })
 
+test_that("Handling deprecation warnings", {
+  resp <- fake_response("https://connect.example/__api__/", headers = list(
+    `X-Deprecated-Endpoint` = "/v1"
+  ))
+  expect_warning(
```
aronatkins, line 46: Do we already confirm that a warning is not generated in the absence of the header?

aronatkins (approved):

R/connect.R
nealrichardson: Sure, why not

tests/testthat/test-connect.R
nealrichardson: If every request raised this warning, the test output would be a mess, and it's not. But I can add an explicit test here.

R/connect.R
nealrichardson: I did this, then pulled on it a little more and refactored the warning code to use the `.frequency` feature built in to rlang, so we could delete some logic here.

PR #269, opened 2024-05-24
Move progress package to Suggests

nealrichardson:
In the spirit of trying to lighten the required dependencies. In a previous PR, I bumped up the default page size to something higher, so there will be fewer requests made that would make a progress bar useful. 

To confirm, I ran tests with and without `progress` installed and they passed. I'm not sure we have a CI job where suggested packages are not installed (though that wouldn't be bad to add).


tests/testthat/test-git.R
```diff
@@ -1,4 +1,6 @@
 without_internet({
+  scoped_experimental_silence()
```
jonkeane, line 2: These are unrelated to this PR and 'just' quieting down tests? Or something else?

R/page.R
```diff
@@ -95,3 +95,15 @@ page_offset <- function(client, req, limit = Inf) {
...
+
+optional_progress_bar <- function(...) {
+  if (requireNamespace("progress", quietly = TRUE)) {
+    progress::progress_bar$new(...)
+  } else {
+    # Return a mock object that behaves enough like a progress bar object
+    list(
+      tick = function() {},
+      finished = TRUE
+    )
```
jonkeane, line 107: What happens without that `finished = TRUE`? I see it referenced at Would https://github.com/rstudio/connectapi/blob/443a2a2049df069a3606459e352569d01e25f06d/R/get.R#L294-L299 But maybe that could be changed / pulled in here and then `finsihed` isn't needed at all?

jonkeane (commented):
A few comments, the `finished = TRUE` looks like it's dangerous (like it'll end quickly when we don't want that to be the case, but maybe it just looks weird.

tests/testthat/test-git.R
nealrichardson: Correct, these test files were added in #264, and I didn't notice them before because I was iterating inside of a live R session and these warnings only fire the first time.

R/page.R
nealrichardson: I'll take a look at refactoring

R/get.R
```diff
@@ -334,13 +328,16 @@ content_list_with_permissions <- function(src, ..., .p = NULL) {
...
   message("Getting permission list")
-  pb <- progress::progress_bar$new(
+  pb <- optional_progress_bar(
     total = nrow(content_list),
     format = "[:bar] :percent :eta"
   )
-  content_list[["permission"]] <- purrr::map(
-    content_list$guid,
-    function(.x) .get_content_permission_with_progress(src, .x, pb)
+  content_list[["permission"]] <- purrr::pmap(
```
nealrichardson, line 335: @jonkeane I refactored this to remove the need for the `finished` field; I also switched this to a `pmap` in order to save a `GET /v1/content/guid/` request for each content item--we already have the data so we don't need to fetch it again. There are surely lots more of these in the code, I'll do some hunting (later).

R/page.R
jonkeane: OOPH sorry about the copy pasta disfluencies there. I had two comments about that chunk elsewhere edited them into one and only did a half good job at that 🤦

R/get.R
jonkeane: Nice!

> There are surely lots more of these in the code, I'll do some hunting (later).

I did search for `.get_content_permission_with_progress` and I thought I only saw it here(!!). But there definitely might be other things that pass the progress bar like this

nealrichardson: Not `.get_content_permission_with_progress`, I mean the pattern of doing `get_content()` to get a data frame of all content, then iterating over `guid` and calling `content_item(connect, guid)`, which does a `GET` to fetch the Content record. But we already have that, every row in that data.frame is Content.

jonkeane: AAAAH, right

nealrichardson:
@jonkeane any other thoughts or is this good to merge?

jonkeane:
Oh, sorry, I didn't realize you were waiting for me! Nope, this is good to go.

Issue #265, opened 2024-05-17
Audit Vanities


kellobri:
The `audit_vanity_urls` function returns all content paths instead of just the vanity paths as one might expect. It looks like this was left as a TODO item in the code: https://github.com/rstudio/connectapi/blob/15f49bcba389b616c4ec51717dca993f65fc24bf/R/audits.R#L42

The code sample in the API reference documentation suggests just calling client$GET on the vanities endpoint. This returns the expected result.

```r
library(connectapi)

# Assumes CONNECT_SERVER and CONNECT_API_KEY are set in the environment

client <- connect()

result <- client$GET("/v1/vanities")
```

nealrichardson:
I don't understand why this function exists. https://github.com/rstudio/connectapi/commit/0c100366e3bf71a45507afb121d09a6b1a7ec1d1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R41-R44 is as best as I can figure.

PR #271, opened 2024-05-24
Switch audit_* functions to use content not "apps"; replace audit_vanity_urls

nealrichardson:
Fixes https://github.com/rstudio/connectapi/issues/265

I dug into this to try to remove usage of the unversioned API in these functions, and along the way I think I made the vanity function do something useful.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -17,3 +17,4 @@
 ^connectapi.*\.tar\.gz$
 ^connectapi.*\.tgz$
 \.orig$
+^.lintr$
--- a/R/connectapi.R
+++ b/R/connectapi.R
@@ -12,6 +12,7 @@ utils::globalVariables(
     "access_type",
     "connectapi_ptypes",
     "guid",
+    "last_deployed_time",
     "owner_guid",
     "parse_connectapi_typed",
     "permission",


R/audits.R
```diff
@@ -1,110 +1,49 @@
...
-  content <- sapply(urls, function(u) {
-    trim_vanity(u, parse_server$path)
-  })
+vanity_is_available <- function(connect, vanity) {
+  current_vanities <- connect$GET(v1_url("vanities"))
+  current_vanity_paths <- purrr::map_chr(current_vanities, "path")
 
-  vanities <- content[!grepl("content/\\d+", content)]
+  # In case a full URL has been given, prune it down to just the path
+  vanity <- sub(paste0("^", connect$server), "", vanity)
```
jonkeane, line 17: This is probably the right approach, but ooph munging URLs like this feels yucky

R/audits.R
```diff
@@ -1,110 +1,49 @@
...
-audit_r_versions <- function(apps) {
-  r_versions <- get_field(apps, "r_version", TRUE)
-  published <- get_field(apps, "last_deployed_time", TRUE)
+audit_r_versions <- function(content) {
+  if (!requireNamespace("ggplot2", quietly = TRUE)) {
+    stop("ggplot2 is required for this function")
+  }
+  if (!requireNamespace("gridExtra", quietly = TRUE)) {
+    stop("gridExtra is required for this function")
+  }
```
jonkeane, line 40: I wonder if we should make this a helper since we are hoping to prune things down more like this. Maybe not here, but seeing them back to back and knowing other places we might use it...

R/audits.R
```diff
@@ -1,110 +1,49 @@
...
-    published = do.call(c, unname( # this flattens the list while preserving the date time
-      lapply(
-        published[!sapply(r_versions, is.null)], # filter out records w/o r version
-        function(d) {
-          lubridate::ymd_hms(d)
-        }
-      ) # convert to date time
-    ))
-  )
+  timeline <- content[!is.na(content$r_version), c("r_version", "last_deployed_time")]
```
jonkeane, line 42: I'm fine to bump the lintr to 100 characters, but it would be good not to make it even more angry than it already is...

R/audits.R
```diff
@@ -114,7 +53,7 @@ audit_r_versions <- function(apps) {
 
   # timeline
   p2 <- ggplot2::ggplot(timeline) +
-    ggplot2::geom_point(pch = 4, ggplot2::aes(x = published, color = r_version, y = r_version)) +
+    ggplot2::geom_point(pch = 4, ggplot2::aes(x = last_deployed_time, color = r_version, y = r_version)) +
```
jonkeane, line 56: Should we add `last_deployed_time` to the global vars list? Or does `vars("last_deployed_time")` work here?

R/audits.R
```diff
@@ -130,66 +69,31 @@ audit_r_versions <- function(apps) {
...
-  } else {
-    run_as_current <- NULL
-  }
-
-
-  return(
-    rbind(run_as, run_as_current)
-  )
+audit_runas <- function(content) {
+  content$run_as <- ifelse(content$run_as_current_user, "current user", content$run_as)
```
jonkeane, line 79: Another line length

tests/testthat/test-audit.R
```diff
@@ -46,3 +12,45 @@ without_internet({
...
+      )
+    )
+  })
+
+  test_that("vanity_is_available", {
+    con <- Connect$new(server = "https://connect.example", api_key = "fake")
+    expect_true(vanity_is_available(con, "not-currently/used"))
+    expect_false(vanity_is_available(con, "/streamlit/my-app/"))
+    expect_false(vanity_is_available(con, "streamlit/my-app"))
+    expect_false(vanity_is_available(con, "https://connect.example/streamlit/my-app"))
```
jonkeane, line 54: 🤯 I don't think I had ever thought that you could even make a "hierarchy" of vanities by `reports/report_one`, `reports/report_two`, and then possibly putting up a thing with connect widgets at `reports` 🤯

jonkeane: I know  ^^^ isn't what you're testing but seeing it here made me realize

R/audits.R
```diff
@@ -1,110 +1,49 @@
...
-  # TODO: why does vanities not work?
-  urls <- get_field(apps, "url")
-  parse_server <- httr::parse_url(server_url)
-  if (is.null(parse_server$scheme)) {
-    stop(glue::glue("ERROR: protocol (i.e. http:// or https://) not defined on server_url={server_url}"))
-  }
-  content <- sapply(urls, function(u) {
-    trim_vanity(u, parse_server$path)
-  })
+vanity_is_available <- function(connect, vanity) {
```
jonkeane, line 12: This might be a server question more than anything, but what happens to characters that need url encoding? Do they come back form the server encoded (such that we should URL encode here before checking availability?)

jonkeane (commented):
Nice clean ups, mostly comments about the angry lintr (IMO we should just bump line length to 100 or so).

One more substantive comment about URL encoding

R/audits.R
nealrichardson: There's probably an rlang helper for this.

R/audits.R
nealrichardson: Good question, I'm not sure.

R/audits.R
nealrichardson: https://rlang.r-lib.org/reference/is_installed.html I guess we could do `rlang::check_installed(c("ggplot2", "gridExtra"))`? The interactive behavior of asking you if you want to install them seems odd to me but maybe that's just me.

R/audits.R
nealrichardson: I added a .lintr file to do that.

R/audits.R
```diff
@@ -1,110 +1,49 @@
...
-        }
-      ) # convert to date time
-    ))
-  )
+  timeline <- content[!is.na(content$r_version), c("r_version", "last_deployed_time")]
 
   # histogram
   p1 <- ggplot2::ggplot(timeline) +
-    ggplot2::geom_histogram(ggplot2::aes(r_version), stat = "count") +
+    ggplot2::geom_bar(ggplot2::aes(r_version)) +
```
nealrichardson, line 46: This is odd, I thought this change was already on main https://github.com/rstudio/connectapi/pull/179/files#diff-7fc2ff5dc6fb4e6e097ab7583ddee9829c37bc07c3cd2f96731b0cd314176d84R107

R/audits.R
nealrichardson: Added to the list

R/audits.R
jonkeane: 🙃 Weird. Maybe stale Github diff? Or it's looking at the branch point and not the merge base?

R/audits.R
nealrichardson: I tried to add a vanity url with spaces in the web app and it errored, so I think this is fine.

image

PR #282, opened 2024-07-11
Remove unnecessary dev dependency from pkgdown.yaml

nealrichardson:
Last CI run on main failed with this: https://github.com/rstudio/connectapi/actions/runs/9898285975/job/27344733776#step:6:1

`git blame` says this is 3 years old, surely the feature has been released by now.


jonkeane (approved):
Thanks for this!

PR #283, opened 2024-07-15
feat:`content_render()` and `content_restart()` implementations

toph-allen:
## Intent

Add `content_render(content)` and `content_restart(content)` functions.

## Approach

- `content_render()` takes a piece of rendered content and triggers a render via the `variants` API. It returns a `ContentTask` object. If called on non-rendered content, raises an error.
- `content_restart()` takes a piece of interactive content and sets then unsets an environment variable. If called on non-interactive content, raises an error.
- Added a `default_variant` property to the `Content` class, an active binding that returns its default variant.
- Added two active bindings (a.k.a. dynamic properties) to `Content`: `is_interactive` and `is_rendered`, both booleans that are computed from `app_mode`.

Added tests for all behaviors. I'm still feeling out testing this library; I feel like I'm used to more fully-specified unit tests rather than the more e2e style tests here.


R/content.R
```diff
@@ -273,6 +273,56 @@ Content <- R6::R6Class(
...
+
+    #' @description Render or restart this content item.
+    refresh = function() {
+      if (is_rendered(self$content$app_mode)) {
+        self$render()
+      } else if (is_interactive(self$content$app_mode)) {
+        self$restart()
+      } else {
+        warning("Content app mode cannot be rendered or restarted.")
+      }
```
jonkeane, line 286: It feels a little duplicative to have this else branch, but then also be checking within `render` and `restart` if they are of the correct type. 

I wonder if it would make sense to either collapse render and refresh here into this one method or to do `self$restart()` in the `else` here and live with the message being slightly off possibly?

I don't have a good answer, but seeing this + the stuff below makes it seem like we might be slightly off on this factoring. Maybe writing tests would help get this factored nicely?

R/content.R
```diff
@@ -273,6 +273,56 @@ Content <- R6::R6Class(
       cat('content_item(client, guid = "', self$get_content()$guid, '")', "\n", sep = "")
       cat("\n")
       invisible(self)
+    },
+
+    #' @description Render or restart this content item.
```
jonkeane, line 278: You don't have to do it in the docstring here, but it's one place to do it. What would it look like to actually use this?

```
connect <- client()
...
...$refresh()
```

R/content.R
```diff
@@ -273,6 +273,56 @@ Content <- R6::R6Class(
...
+    #' @description Restart this application. Only operates on interactive content.
+    restart = function() {
+      if (!is_interactive(self$content$app_mode)) {
+        warning("This content cannot be restarted because it is not an interactive content type.")
+        return(invisible(self))
+      }
+      random_hash = token_hex(32)
+      # https://rlang.r-lib.org/reference/glue-operators.html#what-s-the-deal-with-
+      self$environment_set("{random_hash}" = random_hash)
+      self$environment_set("{random_hash}" = NA)
```
jonkeane, line 314: Is the link to the glue documentation the right anchor? it looks like it's the `:=` operator?

R/utils.R
```diff
@@ -170,3 +170,41 @@ check_connect_version <- function(using_version, minimum_tested_version = "1.8.8
...
+    "quarto-shiny",
+    "tensorflow-saved-model",
+    "api"
+  )
+  if ((app_mode) %in% interactive_app_modes) {
+    TRUE
+  } else {
+    FALSE
+  }
+}
```
jonkeane, line 210: Does the connect API know about the mapping between these two modes?

R/content.R
```diff
@@ -925,3 +975,24 @@ get_content_permissions <- function(content, add_owner = TRUE) {
...
+    variant <- get_variant(content, variant_key)
+  } else {
+    variant <- get_variant_default(content)
+  }
+
+  rendered <- variant$render()
+  rendered$task_id <- rendered$id
+
+  VariantTask$new(connect = variant$get_connect(), content = content$get_content(), key = variant$key, task = rendered)
+}
```
jonkeane, line 998: This only works for rendered content currently, yeah? You're thinking of swapping out some / most of this with the `$refresh()` version up above, or something else?

jonkeane (commented):
I know this is still in progress, but a few comments from me

R/content.R
toph-allen: Yeah, the right anchor is `https://rlang.r-lib.org/reference/glue-operators.html#using-glue-syntax-in-packages`

```suggestion
      # https://rlang.r-lib.org/reference/glue-operators.html#using-glue-syntax-in-packages
      self$environment_set("{random_hash}" = random_hash)
      self$environment_set("{random_hash}" = NA)
```

R/content.R
toph-allen: Yeah, the documentation definitely isn't complete.

R/content.R
toph-allen: Yeah, I'm not sure. This is basically to handle the `static` app-mode, i.e. something that would neither be rendered or restarted. It feels weirdly duplicative with the other checks below, I agree.

I think the weirdness is… maybe largely a symptom of providing `refresh` that does both things and `render` / `restart` that each behave differently.

toph-allen: I'll iterate on this more at the start of next week, to see if something else feels more ergonomically sensible.

R/utils.R
toph-allen: The API doesn't contain data indicating this distinction, if that's what you mean. It feels wrong to have this information spread out like this. Definitely `rendered` vs `server-backed` isn't enough, because you also have static content that is neither. You could argue that we should have a property in the `content` API that indicates whether an app is rendered vs server-backed (i.e. restartable) vs neither.

A close analog to this is the `isWorkerApp` function used in the dashboard: https://github.com/rstudio/connect/blob/1d76aeddb339ec379614f747ee91761eca69b5fc/src/connect/worker/apptypes/apptypes.go#L158-L185

toph-allen: Information on which apps are "worker" app modes _does_ exist in Go: https://github.com/rstudio/connect/blob/1d76aeddb339ec379614f747ee91761eca69b5fc/src/connect/worker/apptypes/apptypes.go#L158-L177

toph-allen: We don't have any equivalent designation for "report" / "rendered" app modes, though, which feels like an oversight.

R/content.R
toph-allen: Yeah, you're right — this was my first pass at using the variant object to do the render, and I took this and updated it for the method implementations above.

We talked a little this week about whether it's better to use methods or functions.

If the methods remain, I'll update this to just call the content item's `$refresh()` method.

R/utils.R
jonkeane: Yeah, we don't have to (and probably shouldn't here), but it might be good to make an issue in the connect repo about what we might do in order to expose this better / more consistently?

tests/testthat/test-content.R
```diff
@@ -180,3 +180,48 @@ without_internet({
     )
   })
 })
+
+with_mock_api({
+  test_that("content_render() calls the correct endpoint, returns task on success", {
+    client <- Connect$new(server = "https://connect.example", api_key = "not-a-key")
+    x <- content_item(client, "951bf3ad-82d0-4bca-bba8-9b27e35c49fa")
```
jonkeane, line 187: This is minor, but this GUID is the full one, but then we return the shortend version in the mocks. Should we also shorten these here too so they are consistent?

tests/testthat/test-content.R
```diff
@@ -180,3 +180,48 @@ without_internet({
...
+    expect_error(content_restart(x), "Restart not supported for application mode: quarto-static. Did you mean content_render()?", fixed = TRUE)
+  })
+})
+
+with_mock_api({
+  test_that("content$default_variant gets the default variant", {
+    client <- Connect$new(server = "http://connect.example", api_key = "not-a-key")
+    x <- content_item(client, "951bf3ad-82d0-4bca-bba8-9b27e35c49fa")
+    v <- x$default_variant # TODO Can't figure out why I can't use `expect_GET`
```
jonkeane, line 223: What's the error you're getting with `expect_GET`?

renv.lock
```diff
@@ -1,6 +1,6 @@
 {
```
jonkeane, line 1: IMHO it would be good to not do all of this renv bumping as part of this work. We might want to bump our pins like this at some point, but the churn in these files is distraction from what we're pushing here

R/utils.R
```diff
@@ -170,3 +170,41 @@ check_connect_version <- function(using_version, minimum_tested_version = "1.8.8
...
```
jonkeane, line 210: We don't have to do this here, but it feels a little funny that we use these at our top level and pass the actual app modes to them. Why not make it a method on a piece of content so that you don't have to know the internal structure of the R6 content method to extract the mode and then pull this out?

jonkeane: And FWIW, I noticed this not in reading these functions themselves but seeing how we call them up in https://github.com/rstudio/connectapi/pull/283/files#diff-2517925fdf51a96dc6acb93b597db70c334538b5b9f033cbe48e1b17a86d3455R954

toph-allen: I thought the exact same thing when I was writing that same code! I'll make that change.

renv.lock
toph-allen: Yes, fair — I had to do this to get the repo running on my machine but should probably cherrypick it into a different PR.

tests/testthat/test-content.R
toph-allen: Yeah, the code in `setup.R` automatically shortens it. I'll shorten it in tests too.

renv.lock
toph-allen: https://github.com/rstudio/connectapi/pull/285

tests/testthat/test-content.R
toph-allen: Most of the other tests use full-length GUIDs. Part of me feels like it's reasonable to leave them full-length because it's easier to visually parse them as a GUID. I'm not sure the reasoning behind truncating them in tests.

tests/testthat/test-content.R
toph-allen: Something like

```
> without_internet({
+   with_mock_api({
+     test_that("content$default_variant gets the default variant", {
+       client <- Connect$new(server = "[https://connect.example](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)", api_key = "not-a-key")
+       x <- content_item(client, "951bf3ad")
+       # v <- x$default_variant # TODO Can't figure out why I can't use `expect_GET`
+       expect_GET(x$default_variant, url="[https://connect.example/__api__/applications/951bf3ad/variants](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)")
+       # expect_identical(v$key, "WrEKKa77")
+     })
+   })
+ })
Defining Connect with server: [https://connect.example](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)
-- Failure: content$default_variant gets the default variant -------------------
`object` did not throw the expected error.
Backtrace:
    x
 1. \-httptest::expect_GET(x$default_variant, url = "[https://connect.example/__api__/applications/951bf3ad/variants](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)")
 2.   \-httptest:::expect_mock_request(object, "GET ", url, " ", ...)
 3.     \-request_happened()(...)

-- Warning: content$default_variant gets the default variant -------------------
Arguments in `...` must be used.
x Problematic arguments:
* fixed = fixed
* ignore.case = ignore.case
* perl = perl
* useBytes = useBytes
i Did you misspell an argument name?
Backtrace:
     x
  1. +-httptest::expect_GET(x$default_variant, url = "[https://connect.example/__api__/applications/951bf3ad/variants](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)")
  2. | \-httptest:::expect_mock_request(object, "GET ", url, " ", ...)
  3. |   \-request_happened()(...)
  4. |     \-testthat:::expect_condition_matching(...)
  5. |       \-rlang (local) `<fn>`()
  6. |         \-rlang:::check_dots(env, error, action, call)
  7. |           \-rlang:::action_dots(...)
  8. |             +-rlang (local) try_dots(...)
  9. |             | \-rlang::try_fetch(expr, error = error)
 10. |             |   +-base::tryCatch(...)
 11. |             |   | \-base (local) tryCatchList(expr, classes, parentenv, handlers)
 12. |             |   |   \-base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
 13. |             |   |     \-base (local) doTryCatch(return(expr), name, parentenv, handler)
 14. |             |   \-base::withCallingHandlers(...)
 15. |             \-rlang (local) action(...)
 16. |               \-rlang:::signal_abort(cnd, .file)
 17. |                 \-base::signalCondition(cnd)
 18. \-rlang (local) `<fn>`(`<rlb_rr__>`)
 19.   \-handlers[[1L]](cnd)

Error:
! Test failed
Hide Traceback
     x
  1. +-httptest::without_internet(...)
  2. | \-base::eval.parent(expr)
  3. |   \-base::eval(expr, p)
  4. +-httptest::with_mock_api(...)
  5. | \-base::eval.parent(expr)
  6. |   \-base::eval(expr, p)
  7. +-testthat::test_that(...)
  8. | \-withr (local) `<fn>`()
  9. \-reporter$stop_if_needed()
 10.   \-rlang::abort("Test failed", call = NULL)
```

toph-allen: I'm sure I'm doing something obviously wrong; I just can't see what yet.

R/content.R
```diff
@@ -274,6 +274,56 @@ Content <- R6::R6Class(
...
+  ),
+  active = list(
+    #' @field default_variant The default variant for this object.
+    default_variant = function(value) {
+      if (missing(value)) {
+        get_variant(self, "default")
+      } else {
+        stop("Default variant cannot be set.", call. = FALSE)
+      }
+    },
```
tdstein, line 286: Why does this function accept `value` if it expects `value` to always be "missing"?  Should it be...
```suggestion
    default_variant = function() {
        get_variant(self, "default")
    },
```

R/content.R
```diff
@@ -925,3 +975,55 @@ get_content_permissions <- function(content, add_owner = TRUE) {
...
+#' Only valid for interactive content (e.g., applications, APIs).
+#' 
+#' @export
+content_restart <- function(content) {
+  validate_R6_class(content, "Content")
+  if (!content$is_interactive) {
+    stop(glue::glue("Restart not supported for application mode: {content$content$app_mode}. Did you mean content_render()?"))
+  }
+  random_hash = token_hex(32)
+  env_var_name = glue::glue("CONNECT_RESTART_{random_hash}")
```
tdstein, line 1024: I used a leading `_` here to provide an extra hint that this variable is "private" by convention.

https://github.com/posit-dev/posit-sdk-py/blob/main/src/posit/connect/content.py#L217C21-L217C42

R/utils.R
```diff
@@ -170,3 +170,11 @@ check_connect_version <- function(using_version, minimum_tested_version = "1.8.8
...
   invisible()
 }
+
+token_hex <- function(n) {
+  raw_to_hex <- function(raw) {
+    paste(as.character(raw), collapse = "")
+  }
+  raw <- as.raw(sample(0:255, n, replace = TRUE))
+  raw_to_hex(raw)
+}
```
tdstein, line 180: Does `raw_to_hex` need to exist as a separate function?

```suggestion
token_hex <- function(n) {
  raw <- as.raw(sample(0:255, n, replace = TRUE))
  paste(as.character(raw), collapse = "")
}
```

tdstein:
It appears the CI failures are due to missing documentation. Once those are added, I think this is ready for final review.

R/content.R
toph-allen: This is an idiosyncrasy of the active properties in `R6`. Basically, the "value missing" case is a getter, the "value present" case is a setter where `value` is the provided value.

The way this looks in R would be:

```r
object$property <- "value"
```

I didn't want to support changing the default variant on the server by modifying an R property.

R/content.R
toph-allen: Ok, I'll do the same. I'm not convinced that we _want_ to use a random hash for the var name here; I think it made something harder in testing, and the orphaned variable in an example I tried looked kinda weird in Connect's UI — I could imagine it making users worry that something is wrong in Connect's dashboard.

But I don't feel that way very strongly _at all_.

How long is the random strong the Python package uses?
```suggestion
  env_var_name = glue::glue("_CONNECT_RESTART_{random_hash}")
```

R/content.R
tdstein: Ah, I see. Thanks for the clarification.

tests/testthat/test-content.R
toph-allen: The error was the same with and without it; that was just the latest iteration. I also tried with the full and shortened GUID.

[P.S. The URLs in the console output are normal-looking URLs, not the weird vscode terminal markdown entities you see here]

```
> with_mock_api({
+   test_that("content$default_variant gets the default variant", {
+     client <- Connect$new(server = "[http://connect.example](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)", api_key = "not-a-key")
+     x <- content_item(client, "951bf3ad-82d0-4bca-bba8-9b27e35c49fa")
+     # v <- x$default_variant # TODO Can't figure out why I can't use `expect_GET`
+     expect_GET(x$default_variant, url="[https://connect.example/__api__/applications/951bf3ad/variants](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)")
+     # expect_GET(x$default_variant, url="[https://connect.example/__api__/applications/951bf3ad-82d0-4bca-bba8-9b27e35c49fa/variants](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)")
+     # expect_identical(v$key, "WrEKKa77")
+   })
+ })
Defining Connect with server: [http://connect.example](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)
-- Failure: content$default_variant gets the default variant -------------------
`object` did not throw the expected error.
Backtrace:
    x
 1. \-httptest::expect_GET(x$default_variant, url = "[https://connect.example/__api__/applications/951bf3ad/variants](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)")
 2.   \-httptest:::expect_mock_request(object, "GET ", url, " ", ...)
 3.     \-request_happened()(...)

-- Warning: content$default_variant gets the default variant -------------------
Arguments in `...` must be used.
x Problematic arguments:
* fixed = fixed
* ignore.case = ignore.case
* perl = perl
* useBytes = useBytes
i Did you misspell an argument name?
Backtrace:
     x
  1. +-httptest::expect_GET(x$default_variant, url = "[https://connect.example/__api__/applications/951bf3ad/variants](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)")
  2. | \-httptest:::expect_mock_request(object, "GET ", url, " ", ...)
  3. |   \-request_happened()(...)
  4. |     \-testthat:::expect_condition_matching(...)
  5. |       \-rlang (local) `<fn>`()
  6. |         \-rlang:::check_dots(env, error, action, call)
  7. |           \-rlang:::action_dots(...)
  8. |             +-rlang (local) try_dots(...)
  9. |             | \-rlang::try_fetch(expr, error = error)
 10. |             |   +-base::tryCatch(...)
 11. |             |   | \-base (local) tryCatchList(expr, classes, parentenv, handlers)
 12. |             |   |   \-base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
 13. |             |   |     \-base (local) doTryCatch(return(expr), name, parentenv, handler)
 14. |             |   \-base::withCallingHandlers(...)
 15. |             \-rlang (local) action(...)
 16. |               \-rlang:::signal_abort(cnd, .file)
 17. |                 \-base::signalCondition(cnd)
 18. \-rlang (local) `<fn>`(`<rlb_rr__>`)
 19.   \-handlers[[1L]](cnd)

Error:
! Test failed
Hide Traceback
    x
 1. +-httptest::with_mock_api(...)
 2. | \-base::eval.parent(expr)
 3. |   \-base::eval(expr, p)
 4. +-testthat::test_that(...)
 5. | \-withr (local) `<fn>`()
 6. \-reporter$stop_if_needed()
 7.   \-rlang::abort("Test failed", call = NULL)
```

R/content.R
tdstein: 32 characters. If the user ever sees it in the UI, then something has gone wrong.

tests/testthat/test-content.R
toph-allen: Talked to @nealrichardson about this. In R `httptest` the expect functions are designed to be used when mocks _aren't_ present. So, we're all good!

R/content.R
nealrichardson: When I've done active bindings in R6 classes before that were not settable, I've done `function()` and it's been fine. What happens if you simplify like @tdstein proposed, does something break?

R/content.R
```diff
@@ -925,3 +975,57 @@ get_content_permissions <- function(content, add_owner = TRUE) {
...
+#' especially if this doesn't happen on a regular schedule.
+#' 
+#' Only valid for rendered content (e.g., most Quarto documents, Jupyter
+#' notebooks, R Markdown reports).
+#' 
+#' @param content The content item you wish to render.
+#' @return A [VariantTask] object that can be used to track completion of the render.
+#' @export
+content_render <- function(content) {
+  suppressWarnings({
```
nealrichardson, line 993: Why are we blanket suppressing warnings?

R/content.R
```diff
@@ -925,3 +975,57 @@ get_content_permissions <- function(content, add_owner = TRUE) {
...
+#' server runs an asynchronous process to render the content. This might be
+#' useful if content needs to be updated after its source data has changed,
+#' especially if this doesn't happen on a regular schedule.
+#' 
+#' Only valid for rendered content (e.g., most Quarto documents, Jupyter
+#' notebooks, R Markdown reports).
+#' 
+#' @param content The content item you wish to render.
+#' @return A [VariantTask] object that can be used to track completion of the render.
+#' @export
```
nealrichardson, line 991: Could you add a (not-run) example that shows how you'd use this, especially with the VariantTask?

R/content.R
```diff
@@ -925,3 +975,57 @@ get_content_permissions <- function(content, add_owner = TRUE) {
...
+content_restart <- function(content) {
+  validate_R6_class(content, "Content")
+  if (!content$is_interactive) {
+    stop(glue::glue("Restart not supported for application mode: {content$content$app_mode}. Did you mean content_render()?"))
+  }
+  random_hash = token_hex(32)
+  env_var_name = glue::glue("_CONNECT_RESTART_{random_hash}")
+  # https://rlang.r-lib.org/reference/glue-operators.html#using-glue-syntax-in-packages
+  content$environment_set("{env_var_name}" := random_hash)
+  content$environment_set("{env_var_name}" := NA)
```
nealrichardson, line 1029: Isn't this sufficient? 

```suggestion
  content$environment_set(`_CONNECT_RESTART` = random_hash)
```

R/content.R
```diff
@@ -925,3 +975,57 @@ get_content_permissions <- function(content, add_owner = TRUE) {
...
+#' Only valid for interactive content (e.g., applications, APIs).
+#' 
+#' @param content The content item you wish to restart.
+#' @export
+content_restart <- function(content) {
+  validate_R6_class(content, "Content")
+  if (!content$is_interactive) {
+    stop(glue::glue("Restart not supported for application mode: {content$content$app_mode}. Did you mean content_render()?"))
+  }
+  random_hash = token_hex(32)
```
nealrichardson, line 1025:
```suggestion
  random_hash <- token_hex(32)
```

R/content.R
```diff
@@ -925,3 +975,57 @@ get_content_permissions <- function(content, add_owner = TRUE) {
...
+#' useful if content needs to be updated after its source data has changed,
+#' especially if this doesn't happen on a regular schedule.
+#' 
+#' Only valid for rendered content (e.g., most Quarto documents, Jupyter
+#' notebooks, R Markdown reports).
+#' 
+#' @param content The content item you wish to render.
+#' @return A [VariantTask] object that can be used to track completion of the render.
+#' @export
+content_render <- function(content) {
```
nealrichardson, line 992: Would it be useful to add `variant` to the signature, like this, and then `get_variant()` in the function? 

```suggestion
content_render <- function(content, variant = "default") {
```

nealrichardson (commented):
A few comments but I didn't get through everything, I can look more later.

tests/testthat/test-content.R
toph-allen: Hmm, I've been finding that leaving as full-length GUIDs means that I can easily recreate a live content item in my local environment, which means that recreating test scenarios or capturing additional responses is much easier.

R/content.R
toph-allen: Oop yeah, I misread the documentation — you can just omit the if/else to get the err-on-assignment behavior. Thanks!

R/content.R
toph-allen: Will do.

R/content.R
toph-allen: Because the variant APIs are experimental and warn the first time they're used in a session.

I'll go through and more tightly scope the warnings though.

R/content.R
toph-allen: I… don't have any objection to this! @tdstein Does Python's `render` function also let you target variants?

R/content.R
toph-allen: No, because that leaves an ugly-looking environment variable in the content environment pane — it'd add another one each time the content was restarted. We need to clean up after ourselves here.

toph-allen: I see two changes:

- Removing the unset. We need that, because otherwise it leaves an ugly-looking environment variable in the content environment pane — it'd add another one each time the content was restarted. We need to clean up after ourselves here.
- I would be fine without adding the hash string to the environment. Here I'm just copying the `posit-sdk-py` implementation. @tdstein what do you think about removing the hash from the var name?

I think I see what you mean — if we just have the same variable name, maybe we don't need to clean up after ourselves, we can just update the variable each time.

I don't think it causes any _harm_ to submit two variable requests — it happens immediately and doesn't cause _extra_ process spins that we'd need to, like, debounce (unless a user manage to refresh their page in between the two requests). Either way, I prefer cleaning up after ourselves. Would be fine removing the hash from the var name (I think I like the cleaner look of that in the dashboard in case orphaned variables do occur. Don't think `_CONNECT_RESTART` is likely to collide with anything.)

R/content.R
toph-allen: Right now this function returns a `ContentTask` type. If this allowed arbitrary variants, it'd have to return a `VariantTask` type till I fix the weird typing around variants. 🤔

tests/testthat/test-content.R
jonkeane: I'm fine leaving them long, but maybe we should _not_ shorten them in `setup.R` then. There is definitely a tension here between making these tests easy/quick to get from a running setup vs. looking nice / easy(er) to digest when coding. But we should be consistent with which we do so that we don't run into a situation where doing something like the following would fail paradoxically:

```
GUID <- "951bf3ad-82d0-4bca-bba8-9b27e35c49fa"
x <- content_item(client, GUID)
expect_true(x$guid, GUID)
```

tests/testthat/test-content.R
```diff
@@ -180,3 +180,44 @@ without_internet({
...
+    expect_error(content_render(x), "Render not supported for application mode: shiny. Did you mean content_restart()?", fixed = TRUE)
+  })
+})
+
+with_mock_api({
+  test_that("content_restart() calls the correct endpoint", {
+    client <- Connect$new(server = "https://connect.example", api_key = "not-a-key")
+    x <- content_item(client, "8f37d6e0-3395-4a2c-aa6a-d7f2fe1babd0")
+    expect_PATCH(content_restart(x), url = "https://connect.example/__api__/v1/content/8f37d6e0/environment")
+    # No returned is returned value; we just test that it made the right calls.
```
jonkeane, line 205: I'm not sure I follow what you mean by "No returned is returned value"

toph-allen: I should probably just delete the comment 😂

toph-allen:
```suggestion
```

R/content.R
nealrichardson: I think there is a `scoped_experimental_silence()` function or something that the tests use, that could be used here.

Or delete the experimental warnings. They're annoying and surely everyone just ignores them.

R/content.R
tdstein: No, the Variants API bindings are not yet public in the Python implementation. But, the eventual implementation will look like:

*via the content resource*

```py
from posit import connect

client = connect.Client()
content = client.content.get(...)
task = content.render(variant_id=...)
```

and 

*via the variant resource*

```py
from posit import connect

client = connect.Client()
content = client.content.get(...)
variant = content.variants.get(...)
task = variant.render()
```

R/content.R
tdstein: I think it will be confusing for users to see the `_CONNECT_RESTART` variable in their envvars list. 

The set/unset routine happens quickly, so it's unlikely a user would see the variable. And as @toph-allen mentioned, Connect denounces the restart side-effect within a window. So, in the rare case that set and unset happen in different windows, it's still very unlikely that a user would be able to load their application successfully before the second window completed.

The `random_hash` is an extra measure to avoid collisions with users, although it's very unlikely a user would have an envvar named  `_CONNECT_RESTART`.

tdstein: However, as I revisit this, I think using a UNIX timestamp would be better than a random hash. In that case, if something did go wrong during the `unset` routine, it's possible to track when the restart was initiated.

tdstein (approved):
This meets our initial design, and the implementation looks correct to me. The outstanding questions do not appear to be blockers, so they can be addressed in subsequent pull requests if necessary.

toph-allen:
@tdstein I'm gonna address a few more comments before I merge; thanks for the approval!

R/content.R
toph-allen: Agreed on `_CONNECT_RESTART_{unix_time_stamp}` :)

R/content.R
jonkeane: > Or delete the experimental warnings. They're annoying and surely everyone just ignores them.

I will second this. I think they are so common with connectapi for doing things that are helpful to people that mot have gotten alarm fatigue and totally ignore them. Especially if scoping the suppression is tedious, better (IMHO) to just remove this one

tests/testthat/test-content.R
nealrichardson: The shortening in setup.R is about just to avoid running into the "long file paths" R CMD check error.

R/content.R
```diff
@@ -274,6 +274,44 @@ Content <- R6::R6Class(
...
+      get_variant(self, "default")
+    },
+
+    #' @field is_rendered TRUE if this is a rendered content type, otherwise FALSE.
+    is_rendered = function() {
+      if ((self$content$app_mode) %in% c("rmd-static", "jupyter-static", "quarto-static")) {
+        TRUE
+      } else {
+        FALSE
+      }
```
nealrichardson, line 290: And similarly below

```suggestion
      self$content$app_mode) %in% c("rmd-static", "jupyter-static", "quarto-static")
```

toph-allen: Of course

R/content.R
toph-allen: Scoping here was very easy, not at all tedious. I'd like to keep this as-is for this PR (I don't want to release an "official" feature that spews out two warnings — there are two without suppressing them!) and remove the warnings in a future PR. I'll create an issue to track.

toph-allen: Oh @nealrichardson you're totally right, `scoped_experimental_silence()` is it.

tests/testthat/test-content.R
toph-allen: Unsure how to resolve this; I'm going to merge and leave this as tension to resolve in the future. It's somewhat confusing, but not unmanageably so. Leaving this conversation unresolved.

PR #285, opened 2024-07-30
chore: update renv environment and infra

toph-allen:
No major changes; just a single commit cherry-picked from #283.

### DIFF
--- a/renv.lock
+++ b/renv.lock
@@ -1,6 +1,6 @@
 {
   "R": {
-    "Version": "4.2.1",
+    "Version": "4.4.1",
     "Repositories": [
       {
         "Name": "CRAN",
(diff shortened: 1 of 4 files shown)


tdstein:
Are we confident automated testing will catch any potential issues when upgrading dependencies?

nealrichardson:
https://github.com/r-lib/actions/tree/v2-branch/setup-r-dependencies installs dependencies based on the package `DESCRIPTION` file, so this isn't relevant for CI.

tdstein (approved):

jonkeane:
I'm fine with keeping renv around like this, but it feels like a lot of churn + constraint on developers (they must have precisely the same versions of everything) than is helpful. Is it worth writing up an issue for discussion of if that is something we want to maintain?

PR #291, opened 2024-08-27
fix: parsing time stamps with time zone offsets containing colons

### LINKED ISSUE #290: `get_audit_logs` breaks time
It seems that date time format has changed for audit logs and now time column is empty aka `NA`

problematic line:
https://github.com/rstudio/connectapi/blob/f053cff40b55dc46584a141d5458bef7e72cc52f/R/parse.R#L111

example:
```r
as.POSIXct(strptime('2023-08-22T14:13:14+03:00', format = "%Y-%m-%dT%H:%M:%SZ"))
```

toph-allen:
Fixes #290

## Intent

Fix a bug involving the handling of timestamps from Connect.

## Approach

Connect sends timestamps in RFC 3339 format. Specifically:

- Timestamps with no UTC offset end in `Z`, e.g. `"2020-01-01T01:02:03Z"`
- Timestamps with an offset end in `±hh:mm`, e.g. `"2023-08-22T15:13:14+01:00"`

Details on the approaches used.

- This PR fixes a bug where we would not parse time stamps with an offset, resulting in NA values. It uses a combination of regular expressions and using different time format specifiers for different incoming time stamps.
- Converts all incoming timestamps to UTC. Previously, we only parsed UTC times at all, but just fixing the time format revealed a bug where we assigned UTC to non-UTC timestamps.
- Fixes a bug in `make_timestamp()`, which creates timestamps for Connect. Previously, would output all timestamps as UTC, even though they were in the local system time.

## Tests

Added tests for a range of expected inputs. Tests take place in multiple time zones.

### DIFF
--- a/R/parse.R
+++ b/R/parse.R
@@ -123,15 +123,15 @@ coerce_datetime <- function(x, to, ...) {
   }
 }
 
-# Parse character dates received from Connect which use RFC 3339.
+# Parses a character vector of dates received from Connect, using use RFC 3339,
+# returning a vector of POSIXct datetimes.
 #
-# R parses as ISO 8601. When specifying %z, it expects time zones to be
-# specified as `-1400` to `+1400`.
-# Connect returns times in RFC 3339. It denotes time zones with `-14:00` to
-# `+14:00`, and indicates zero offset with `Z`.
+# R parses character timestamps as ISO 8601. When specifying %z, it expects time
+# zones to be specified as `-1400` to `+1400`.
+#
+# Connect's API sends times in a specific RFC 3339 format: indicating time zone
+# offsets with `-14:00` to `+14:00`, and zero offset with `Z`.
 # https://github.com/golang/go/blob/54fe0fd43fcf8609666c16ae6d15ed92873b1564/src/time/format.go#L86
-# I don't understand why replacing Z with `+0000` and parsing with %z doesn't work, but it doesn't.
-# We have to parse zero-offset time stamps expecting literal Zs.
 # For example:
 # - "2023-08-22T14:13:14Z"
 # - "2023-08-22T15:13:14+01:00"
@@ -139,15 +139,11 @@ coerce_datetime <- function(x, to, ...) {
 parse_connect_rfc3339 <- function(x) {
   # Convert any timestamps with offsets to a format recognized by `strptime`.
   x <- gsub("([+-]\\d\\d):(\\d\\d)$", "\\1\\2", x)
-  
-  # Times with and without offsets require different formats, so create a vector
-  # of formats to be used in parallel.
-  format_strings <- ifelse(
-    grepl("Z$", x),
-    "%Y-%m-%dT%H:%M:%SZ",
-    "%Y-%m-%dT%H:%M:%S%z"
-  )
 
+  # `purrr::map2_vec()` converts to POSIXct automatically, but we need
+  # `as.POSIXct()` in there to account vectors of length 1, which it seems are
+  # not converted.
+  # 
   # Parse with an inner call to `strptime()`; convert the resulting `POSIXlt`
   # object to `POSIXct`.
   #
@@ -161,11 +157,14 @@ parse_connect_rfc3339 <- function(x) {
   # [1] "America/New_York"
   # > as.POSIXct(xlt, tz = "UTC")
   # [1] "2024-08-29 16:36:33 UTC"
-  #
-  # `purrr::map2_vec()` converts to POSIXct automatically, but we need
-  # `as.POSIXct()` in there to account for only one item.
-  purrr::map2_vec(x, format_strings, function(.x, .y) {
-    as.POSIXct(strptime(.x, format = .y, tz = "UTC"))
+  purrr::map_vec(x, function(.x) {
+  # Times with and without offsets require different formats.
+    format_string = ifelse(
+      grepl("Z$", .x),
+      "%Y-%m-%dT%H:%M:%SZ",
+      "%Y-%m-%dT%H:%M:%S%z"
+    )
+    as.POSIXct(strptime(.x, format = format_string, tz = "UTC"))
   })
 }
 
--- a/R/parse.R
+++ b/R/parse.R
@@ -18,7 +18,13 @@ make_timestamp <- function(input) {
     # TODO: make sure this is the right timestamp format
     return(input)
   }
-  safe_format(input, "%Y-%m-%dT%H:%M:%SZ")
+  
+  # In the call to `safe_format`:
+  # - The format specifier adds a literal "Z" to the end of the timestamp, which
+  #   tells Connect "This is UTC".
+  # - The `tz` argument tells R to produce times in the UTC time zone.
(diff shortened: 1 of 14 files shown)


tdstein (approved):
Nicely done! Thanks for all of the comments. Do we need to verify that dates roll over correctly from different time zones into UTC?

R/parse.R
```diff
@@ -118,6 +123,52 @@ coerce_datetime <- function(x, to, ...) {
...
+  # > xlt
+  # [1] "2024-08-29 16:36:33 EDT"
+  # > tzone(xlt)
+  # [1] "America/New_York"
+  # > as.POSIXct(xlt, tz = "UTC")
+  # [1] "2024-08-29 16:36:33 UTC"
+  #
+  # `purrr::map2_vec()` converts to POSIXct automatically, but we need
+  # `as.POSIXct()` in there to account for only one item.
+  purrr::map2_vec(x, format_strings, function(.x, .y) {
```
aronatkins, line 167: Is the `function(.x, .y)` called on each element? That gives you the option of moving format string selection into this function call rather than computing it separately.

I'm fine leaving this as-is; your choice about which option is more clear.

tests/testthat/test-parse.R
```diff
@@ -25,13 +25,67 @@ test_that("coerce_datetime fills the void", {
   expect_error(coerce_datetime(NA_complex_, NA_datetime_, name = "complexity"), class = "vctrs_error_incompatible_type")
 })
 
-test_that("make_timestamp works with POSIXct", {
-  outcome <- "2020-01-01T01:02:03Z"
+test_that("parse_connect_rfc3339 parses timestamps we expect from Connect", {
+  original_tz <- Sys.getenv("TZ")
+  withr::defer(Sys.setenv(TZ = original_tz))
```
aronatkins, line 30: Alternatively:

```r
withr::local_envvar(TZ = Sys.getenv("TZ"))
```

tests/testthat/test-parse.R
```diff
@@ -25,13 +25,67 @@ test_that("coerce_datetime fills the void", {
...
+  )
+
+  expected <- as.POSIXct(strptime(c(
+    "2023-08-22T14:13:14+0000",
+    "2020-01-01T01:02:03+0000",
+    "2023-08-22T15:13:14+0100",
+    "2020-01-01T00:02:03-0100"
+  ), format = "%Y-%m-%dT%H:%M:%S%z", tz = "UTC"))
+
+  Sys.setenv(TZ = "America/New_York")
```
aronatkins, line 46: nit: consider adding a timezone on the other side of UTC.

aronatkins (approved):
Lacks NEWS update. Approval assumes addition of NEWS.

R/parse.R
nealrichardson: Is it actually possible for the format string to vary by element (e.g., a creation timestamp in user records would be omitted as UTC for some users and local time for others)? My understanding of the issue was that it was about the configuration of the Connect server.

If it is the case that the Connect server is consistent with itself, then we can determine the format string to use from the first element, and we won't need purrr at all down here, just strptime.

tests/testthat/test-parse.R
```diff
@@ -25,13 +25,67 @@ test_that("coerce_datetime fills the void", {
   expect_error(coerce_datetime(NA_complex_, NA_datetime_, name = "complexity"), class = "vctrs_error_incompatible_type")
 })
 
-test_that("make_timestamp works with POSIXct", {
-  outcome <- "2020-01-01T01:02:03Z"
+test_that("parse_connect_rfc3339 parses timestamps we expect from Connect", {
+  original_tz <- Sys.getenv("TZ")
+  withr::defer(Sys.setenv(TZ = original_tz))
+
+  xs <- c(
```
nealrichardson, line 32: If my suggestion above is correct, then you'd need to revise these tests too, IIUC.

R/parse.R
toph-allen: @aronatkins Yes, it is called on each element. In that case, I could just use `map_vec` and not bother with the parallel creation.

@nealrichardson Yeah, I think you're right, I don't think that a given Connect server will emit multiple records with different time zone offsets, so we can probably determine the format with the first element and use a single format specification.

tests/testthat/test-parse.R
toph-allen: I was gonna say, yes, the tests will need to be reworked, but that's doable.

R/parse.R
toph-allen: @nealrichardson I think I would prefer to keep a function such that it's vectorized in a way that can parse an character of strings that are in the assumed format, rather than introduce the additional assumption that the first element's time zone offset is representative of all of the elements. I know this function is a single-purpose one, but we don't have anything in place that's guaranteeing that all the time zone offsets in the incoming string are the same (other than, like, we write the Connect server).

I guess I'd rather write something that's robust to misuse or unexpected edge cases than something which is marginally more performant but more brittle.

nealrichardson: > we don't have anything in place that's guaranteeing that all the time zone offsets in the incoming string are the same (other than, like, we write the Connect server)

You're right that the performance impact is negligible--any time we're fetching things over HTTP, it's not worth worrying about microseconds of optimization in the client code.

I would counter, as a general rule, that we should avoid the tendency of overgeneralizing and flexibilizing our client code. There's already enough flex we have to add to support different versions of Connect, etc., we don't need more. This to me feels like it should be a design constraint in the Connect API: all timestamps are emitted the same way. (Ideally, Connect would be even stricter, always emitting UTC, ISO 8601, etc.) And thus API clients can assume that timestamps are a certain way. 

In practice here, it's not a big deal--the code that makes this flexible is probably simpler. Just wanted to make the more general design point.

tests/testthat/test-parse.R
nealrichardson: As I said above, I'm ok with leaving the more flexible code in place that you have, but let's do rewrite the tests so that they reflect the behavior that Connect has, i.e. we don't have heterogeneous timestamp formatting in a single vector. That's not a behavior we need to support, so let's not write tests that codify that as behavior that is important, so that whenever we refactor this in the future, we don't imagine that to be a feature requirement.

R/parse.R
toph-allen: Thanks -- I definitely understand the need to not over-generalize, and it's good to keep that in mind. I think this code is striking the right balance — it's pretty narrow in what it expects (only parsing timestamps in the subset of RFC 3339 which includes time zone offsets ending in `Z` / `±hh:mm`).

It's a moot point now, but I realize that another reason constraining it further feels wrong is that it breaks with expectations of how vectorization works in R: functions that operate on vectors generally don't alter processing of subsequent items based on the first item. Better to not break those assumptions IMO.

tests/testthat/test-parse.R
toph-allen: Tests now include single-time-zone vectors with multiple and single elements, offsets and zero offsets.

nealrichardson (approved):

PR #293, opened 2024-08-30
chore: add pull request template

toph-allen:
Mostly added for the checklist. Maybe a bit overkill.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+> If applicable:
+> - Include guidance for reviewers (rewrites ..., start with ..., commit-by-commit, etc.).
+> - Link to notes and discussions about this work.
+
+## Automated Testing
+
+> Describe the automated testing of this change.
+>
+> - What cases are covered by unit tests?
+> - What cases are covered by integration tests?
```
tdstein, line 23: I don't think this is necessary since any changes should be tested.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
+## Intent
+
+> Describe what problem this pull request resolves, or what new feature it
+> introduces.
+>
+> - Link to related GitHub issues, if applicable
```
tdstein, line 6: If you use HTML-style comments, then these notes won't render as part of the description.

tdstein (approved):

.github/PULL_REQUEST_TEMPLATE.md
toph-allen: Yeah, sounds good.

.github/PULL_REQUEST_TEMPLATE.md
toph-allen: I kinda like optimizing for the editing view, though, since that's where folks will interact with these most of the time.

PR #302, opened 2024-09-06
feat: new function `set_content_image()`, deprecate old `set_image_*` functions

### LINKED ISSUE #294: simplify `set_image_*()` functions
There are three functions to update content images:

- `set_image_path(content, path)` uploads a local image file to Connect for use as a content image;
- `set_image_url(content, url)` downloads an image from a URL and calls `set_image_path()`;
- `set_image_webshot(content, ...)` uses `webshot2` to take a screenshot of the content, and calls `set_image_path()` with that resulting image.

The first function should just become `set_content_image(content, path)` or something similar. The other two should become recipes in the cookbook.

The tests for `set_image_webshot()` are very flaky, so they have been disabled in CI.

toph-allen:
Fixes #294

- Deprecate `set_image_path(content, path)`, `set_image_url(content, url)`, `set_image_webshot(content)`.
- Add a new function, `set_content_image(content, path)`. The new function can handle a local file path or a remote URL. If the path points to a local file that exists, it will use that. Otherwise, if it is an `https` or `http` URL, it will attempt to download that image to a temporary file and use that (the old behavior of `set_image_url()`.

In a future PR to the `connect` repo, `set_image_webshot()` will become a recipe in the SDK Cookbook.

Note: this wasn't based off the 0.3.0 release, so the NEWS will have a merge conflict once that PR has merged. Not sure why I started it from `main`, but I did 🤷🏻.


toph-allen:
Oops, accidentally closed with a force push.

PR #303, opened 2024-09-06
feat: new function `set_content_image()`, deprecate old `set_image_*` functions

### LINKED ISSUE #294: simplify `set_image_*()` functions

### LINKED ISSUE #304: use v1 content image API (2024.09.0)
The Connect 2024.09.0 release publicizes content image endpoints.

* `GET /content/{guid}/__thumbnail__`
* `PUT /__api__/v1/content/:guid/thumbnail`
* `DELETE /__api__/v1/content/:guid/thumbnail`

The connectapi functions to get, set, and delete content images need to prefer these public API paths.

https://github.com/rstudio/connectapi/blob/bfefb00ce4721c0bb80aedf58ae217ff63742bee/R/deploy.R#L423-L426
https://github.com/rstudio/connectapi/blob/bfefb00ce4721c0bb80aedf58ae217ff63742bee/R/deploy.R#L479

https://github.com/rstudio/connectapi/blob/bfefb00ce4721c0bb80aedf58ae217ff63742bee/R/deploy.R#L509-L512

https://github.com/rstudio/connectapi/blob/bfefb00ce4721c0bb80aedf58ae217ff63742bee/R/deploy.R#L465

```[tasklist]
### Tasks
```

toph-allen:
Fixes #294
Fixes #304

- Deprecate `set_image_path(content, path)`, `set_image_url(content, url)`, `set_image_webshot(content)`.
- Add a new function, `set_content_image(content, path)`. The new function can handle a local file path or a remote URL. If the path points to a local file that exists, it will use that. Otherwise, if it is an `https` or `http` URL, it will attempt to download that image to a temporary file and use that (the old behavior of `set_image_url()`.

In a future PR to the `connect` repo, `set_image_webshot()` will become a recipe in the SDK Cookbook.

Note: this wasn't based off the 0.3.0 release, so the NEWS will have a merge conflict once that PR has merged. Not sure why I started it from `main`, but I did 🤷🏻.

Replaces #302, which I accidentally closed with a force push.


NEWS.md
```diff
@@ -1,5 +1,12 @@
 # Unreleased
 
+## Breaking changes
+
+- `set_image_path()`, `set_image_url()`, and `set_image_webshot()` have been
+  deprecated and will be removed in a future update. They have been replaced by
+  `set_content_image(content, path)`, which works both with local file paths and
+  remote URLs to images.
```
jonkeane, line 8: It might be worth adding in here that `_webshot` is not being replaced, instead see the recipe at: ...

R/deploy.R
```diff
@@ -487,49 +487,65 @@ has_image <- function(content) {
...
+#' @param path A file path or URL to an image
 #'
-#' @rdname set_image
 #' @family content functions
+#' @rdname set_content_image
 #' @export
-set_image_path <- function(content, path) {
-  warn_experimental("set_image_path")
+set_content_image <- function(content, path) {
+  warn_experimental("set_content_image")
```
jonkeane, line 497: Does this need to be experimental?

R/deploy.R
```diff
@@ -487,49 +487,65 @@ has_image <- function(content) {
...
+  } else {
+    parsed <- httr::parse_url(path)
+    if (parsed$scheme %in% c("http", "https")) {
+      valid_path <- fs::file_temp(pattern = "image", ext = fs::path_ext(parsed[["path"]]))
+      httr::GET(url, httr::write_disk(valid_path))
+      on.exit(unlink(valid_path))
+    }
+  }
+  if (is.null(valid_path)) {
+    stop("Could not locate image at `path`")
```
jonkeane, line 512: Would it be helpful to put the actual path in this message too?

R/deploy.R
```diff
@@ -487,49 +487,65 @@ has_image <- function(content) {
...
   validate_R6_class(content, "Content")
-  guid <- content$get_content()$guid
 
+  valid_path <- NULL
+  if (file.exists(path)) {
+    valid_path <- path
+  } else {
+    parsed <- httr::parse_url(path)
+    if (parsed$scheme %in% c("http", "https")) {
+      valid_path <- fs::file_temp(pattern = "image", ext = fs::path_ext(parsed[["path"]]))
```
jonkeane, line 506: We might use `fs` elsewhere, but I wonder if we need to use it at all here as opposed to `tempfile()`? Do we need to maintain the extension? It's a tempfile anyway so for local filesystem reasons we don't, but it's possible it's required on connect?

R/deploy.R
```diff
@@ -487,49 +487,65 @@ has_image <- function(content) {
...
-#' @param ... Additional arguments passed on to [webshot2::webshot()]
+#' @param path A file path or URL to an image
 #'
-#' @rdname set_image
 #' @family content functions
+#' @rdname set_content_image
 #' @export
-set_image_path <- function(content, path) {
-  warn_experimental("set_image_path")
+set_content_image <- function(content, path) {
```
aronatkins, line 496: Could we treat `path` similarly to `read.csv()`, `readr::read_csv()`, and other functions, where the `file` argument could be a connection as well as a file name or URL?

Consider when the caller already has image data in memory.

R/deploy.R
toph-allen: Yeah, `fs` was in here before, but we don't need it for temporary files.

R/deploy.R
toph-allen: Of course 😬

NEWS.md
toph-allen: Yeah, I'll update this PR when I have created a `connect` PR with the recipe URL, which I'll do before this merges.

R/deploy.R
toph-allen: The package's previous policy seems to have been to use this warning for all unversioned URLs, which this function still uses. I don't feel strongly about keeping the warning; however, changing it opens questions in my mind about all the other places it's used.

I guess, it doesn't seem like the `v1` / unversioned nature of our API actually corresponds to "experimental" / "stable", and definitely not for user-facing functions. And since we're planning to make all unversioned APIs versioned, perhaps all of these warnings can go away?

R/deploy.R
toph-allen: This seems sensible.

NEWS.md
toph-allen: Hmm, if I link to the recipe, would this need to wait until Connect is released? 🤔

R/deploy.R
aronatkins: https://github.com/rstudio/connectapi/issues/304 - I neglected to create a connectapi tracking issue previously.

R/deploy.R
```diff
@@ -487,49 +487,65 @@ has_image <- function(content) {
...
+      valid_path <- fs::file_temp(pattern = "image", ext = fs::path_ext(parsed[["path"]]))
+      httr::GET(url, httr::write_disk(valid_path))
+      on.exit(unlink(valid_path))
+    }
+  }
+  if (is.null(valid_path)) {
+    stop("Could not locate image at `path`")
+  }
+
+  guid <- content$get_content()$guid
```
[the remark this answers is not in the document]
toph-allen: Part of me wonders if the only reason we have `$get_content()` is that `content$content` looks really weird. 😂

toph-allen:
Added first pass at supporting the v1 URLs. The exact conditional flow differs between different uses, because some of the calls include parsing the `httr` response object, but not all.

tests/integrated/test-deploy.R
```diff
@@ -165,11 +165,11 @@ test_that("deploy_current works", {
 
 # image ---------------------------------------------------
 
-test_that("set_image_path works", {
+test_that("set_content_image works with local images", {
```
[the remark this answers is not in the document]
toph-allen: Yeah, let's chat about this tomorrow. Definitely agree that we should have some mock tests, *especially* with supporting the v1 and unversioned URLs — those are new patterns.

R/deploy.R
```diff
@@ -448,89 +456,157 @@ get_image <- function(content, path = NULL) {
...
+delete_thumbnail <- function(content, path = NULL) {
   validate_R6_class(content, "Content")
   guid <- content$get_content()$guid
 
   con <- content$get_connect()
 
   if (!is.null(path)) {
     scoped_experimental_silence()
-    get_image(content, path)
+    get_thumbnail(content, path)
```
nealrichardson, line 476: Maybe the function names are just confusing to me, but why does `delete_thumbnail()` have to `get` first? Why is there a `path` argument to delete?

README.md
```diff
@@ -107,15 +107,15 @@ content <- client %>%
 # set an image for content
 
 content %>%
-  set_image_path("./my/local/image.png")
+  set_content_image("./my/local/image.png")
```
nealrichardson, line 110: Since you've renamed the other `$verb_image()` functions to `$verb_thumbnail()`, should this be `set_content_thumbnail()`?

R/deploy.R
```diff
@@ -400,30 +400,38 @@ deploy_current <- function(content) {
...
   validate_R6_class(content, "Content")
   guid <- content$get_content()$guid
 
   con <- content$get_connect()
 
+  # Connect 2024.09.0 introduced public endpoints for content thumbnails. We
+  # prefer those, falling back to the unversioned endpoints if unavailable.
   res <- con$GET(
-    unversioned_url("applications", guid, "image"),
+    path = NA, # `path` is used only if `url` is not provided.
```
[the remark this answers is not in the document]
toph-allen: The `path` argument has no default value, so it's required, but it isn't used it you pass `URL`, and that doesn't feel intuitive, so I left a message to future folks working with the code.

R/deploy.R
toph-allen: The way it has existed previously, `delete_thumbnail()` saves the image to `path` if `path` is provided before deleting. I've tried to scope the changes in this PR to renaming / deprecating, not also changing the functionality.

toph-allen: Another option would be to keep the old implementation as `delete_image()` and have the new `delete_thumbnail()` _only_ delete, not also get.

nealrichardson: > Another option would be to keep the old implementation as the deprecated `delete_image()` and have the new `delete_thumbnail()` _only_ delete, not also get.

Yeah I like that 🚮

R/deploy.R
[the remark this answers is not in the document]
nealrichardson: It's not required: it's never accessed in the function if you provide `url`, so it doesn't matter whether it's defined or not. Like this:

```
> f <- function(a, b = a + 1) b 
> f(1)
[1] 2
> f(b = 1)
[1] 1
```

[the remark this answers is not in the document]
toph-allen: Oh right, thanks.

R/deploy.R
toph-allen: Like, users can `GET` before they `DELETE` if they want, we're only saving people one line of code in an uncommon edge case.

nealrichardson: 💯

NEWS.md
```diff
@@ -1,3 +1,12 @@
+# Unreleased
+
+## Breaking changes
+
+- `set_image_path()`, `set_image_url()`, have been deprecated and will be
+  removed in a future update. `set_image_webshot()` has been moved to the
+  Connect SDK Cookbook. They have been replaced by `set_content_image(content,
+  path)`, which works both with local file paths and remote URLs to images.
```
nealrichardson, line 8:
```suggestion
  Connect SDK Cookbook. They have been replaced by `set_thumbnail()`,
  which works both with local file paths and remote URLs to images.
```

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+  validate_R6_class(content, "Content")
+  guid <- content$get_content()$guid
+
+  con <- content$get_connect()
+
+  # Connect 2024.09.0 introduced public endpoints for content thumbnails. We
+  # prefer those, falling back to the unversioned endpoints if unavailable.
+  res <- con$GET(
+    url = con$server_url("content", guid, "__thumbnail__"),
+    parser = NULL
```
nealrichardson, line 23: If the goal is to end up with the content written to a file, why not use `httr::write_disk()` here?

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+#' `delete_thumbnail` removes the thumbnail (optionally saving to the given path)
+#' `has_thumbnail` returns whether the content has an thumbnail
+#'
+#' @param content A content object
+#' @param path optional. The path to the image on disk
+#'
+#' @rdname get_thumbnail
+#' @family content functions
+#' @export
+get_thumbnail <- function(content, path = NULL) {
```
nealrichardson, line 13: I don't understand the use case for `path = NULL`. What if `path` were a required argument, and if someone wanted to write to a temp file, they can explicitly say `path = tempfile()`. That's better anyway so that the caller can manage the lifetime of that temp file (as we should in tests).

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+  )
+  if (httr::status_code(res) == 404) {
+    res <- con$GET(
+      unversioned_url("applications", guid, "image"),
+      parser = NULL
+    )
+  }
+
+  if (httr::status_code(res) == 204) {
+    return(NA)
```
nealrichardson, line 33: This strikes me as odd behavior: sometimes the function returns `path`, sometimes `NA`. But at least if we're going to do that, let's return a string NA so that the function is type-stable--otherwise you couldn't `map_chr()` over this.

```suggestion
    return(NA_character_)
```

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+    } else {
+      # try png
+      warning(glue::glue("Could not infer file extension from content type: {ct}. Using '.png'"))
+      path <- fs::file_temp(pattern = "content_image_", ext = ".png")
+    }
+  }
+
+  writeBin(httr::content(res, as = "raw"), path)
+
+  return(fs::as_fs_path(path))
```
nealrichardson, line 52: I'd leave it to the caller to `as_fs_path()` if they want

```suggestion
  path
```

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+      unversioned_url("applications", guid, "image"),
+      parser = NULL
+    )
+  }
+
+  if (httr::status_code(res) == 204) {
+    return(NA)
+  }
+
+  # guess file extension
```
nealrichardson, line 36: If you take my suggestion above and make `path` required, we can delete this whole block. Otherwise, we shouldn't have to guess the file extension--we should know. There's probably a finite set of Content-Types that the Connect API will emit. There's also [this map](https://github.com/yihui/mime/blob/main/R/mimemap.R) of content types to extensions we could reach to if we wanted--`httr` imports `mime` so we already have it available.

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  writeBin(httr::content(res, as = "raw"), path)
+
+  return(fs::as_fs_path(path))
+}
+
+#' @rdname get_thumbnail
+#' @export
+get_image <- function(content, path = NULL) {
```
nealrichardson, line 57: If you were to make `path` required in `get_thumbnail()`, you'd want to be sure not to send `NULL` from here.

```suggestion
get_image <- function(content, path = tempfile()) {
```

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+#' @rdname get_thumbnail
+#' @export
+delete_thumbnail <- function(content, path = NULL) {
+  validate_R6_class(content, "Content")
+  guid <- content$get_content()$guid
+
+  con <- content$get_connect()
+
+  if (!is.null(path)) {
+    scoped_experimental_silence()
```
nealrichardson, line 71: No longer needed right?

```suggestion
```

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+
+#' @rdname get_thumbnail
+#' @export
+delete_thumbnail <- function(content, path = NULL) {
+  validate_R6_class(content, "Content")
+  guid <- content$get_content()$guid
+
+  con <- content$get_connect()
+
+  if (!is.null(path)) {
```
nealrichardson, line 70: I would move this case into `delete_image()` for backwards compatibility, and then `delete_thumbnail` can just be `function(content)`.

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+  )
+  if (httr::status_code(res) == 404) {
+    res <- con$DELETE(
+      unversioned_url("applications", guid, "image"),
+      parser = NULL
+    )
+  }
+  con$raise_error(res)
+
+  return(content)
```
nealrichardson, line 89:
```suggestion
  invisible(content)
```

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+set_thumbnail <- function(content, path) {
+  validate_R6_class(content, "Content")
+
+  valid_path <- NULL
+  if (file.exists(path)) {
+    valid_path <- path
+  } else {
+    parsed <- httr::parse_url(path)
+    if (parsed$scheme %in% c("http", "https")) {
+      valid_path <- fs::file_temp(pattern = "image", ext = fs::path_ext(parsed[["path"]]))
```
nealrichardson, line 149:
```suggestion
      valid_path <- tempfile()
```

R/thumbnail.R
```diff
@@ -0,0 +1,233 @@
...
+      httr::GET(url, httr::write_disk(valid_path))
```
nealrichardson, line 150: What happens if this returns 404?

toph-allen:
Back from vacation and assessing the scope of this PR, I'm going to hold off on implementing changes to the `Connect` class to do with server settings, checking the Connect version, memoizing failed requests to new endpoint versions; those will come in a future pull request.

R/thumbnail.R
toph-allen: Bad stuff — it just sets the image to (I assume) the 404 response, interpreted as a binary image. Thanks.

R/thumbnail.R
toph-allen: Going to take your suggestion and make `path` required — people can pass in `tempfile()` if needed. I'm moving the old code to the old `get_image()` function so that still behaves as before.

README.Rmd
```diff
@@ -111,15 +111,15 @@ content <- client %>%
 # set an image for content
 
 content %>%
-  set_image_path("./my/local/image.png")
+  set_content_image("./my/local/image.png")
```
nealrichardson, line 114: set_thumbnail? I don't see this function name in the diff

R/thumbnail.R
toph-allen: Closing the loop here: because we want to get the result before deciding the file path, so we can determine the path to write to (see https://github.com/rstudio/connectapi/pull/303#discussion_r1775910845).

R/thumbnail.R
toph-allen: The workflow without `NULL` files is janky if you don't know the format of the image you are about to receive.

jonkeane (commented):
Ack, meant to send these this morning, but apparently hadn't pressed the button

tests/integrated/test-thumbnail.R
```diff
@@ -0,0 +1,88 @@
...
+})
+
+test_that("set_thumbnail works with remote paths", {
+  scoped_experimental_silence()
+
+  res <- set_thumbnail(cont1_content, glue::glue("{cont1_content$get_connect()$server}/connect/__favicon__"))
+
+  expect_true(validate_R6_class(res, "Content"))
+
+  # TODO: verify round-trip on the image is actually correct... SHA?
```
jonkeane, line 87: I wonder if we need to do this at all? It's more or less testing that the connect server receives it and doesn't mess up from there — we have server /API tests for that, right?

tests/integrated/test-thumbnail.R
```diff
@@ -0,0 +1,88 @@
+test_that("set_thumbnail works with local images", {
+  scoped_experimental_silence()
+  img_path <- rprojroot::find_package_root_file("tests/testthat/examples/logo.png")
```
jonkeane, line 3:
```suggestion
  img_path <- "examples/logo.png"
```

Does this not work?

R/thumbnail.R
```diff
@@ -38,17 +39,18 @@ get_thumbnail <- function(content, path = NULL) {
...
     } else {
       # Try png
       warning(glue::glue("Could not infer file extension from content type: {ct}. Using '.png'"))
       path <- tempfile(pattern = "content_image_", fileext = ".png")
     }
   }
 
-  writeBin(httr::content(res, as = "raw"), path)
+  contents <- httr::content(res, as = "raw")
+  writeBin(contents, path)
```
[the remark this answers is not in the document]
toph-allen: No, there isn't; this is ugly. I'm just gonna make it a one-liner.

tests/integrated/test-thumbnail.R
```diff
@@ -0,0 +1,88 @@
...
+  res <- set_thumbnail(cont1_content, img_path)
+
+  expect_true(validate_R6_class(res, "Content"))
+})
+
+test_that("get_thumbnail works", {
+  scoped_experimental_silence()
+  img_path <- rprojroot::find_package_root_file("tests/testthat/examples/logo.png")
+
+  tmp_img <- fs::file_temp(pattern = "img", ext = ".png")
```
jonkeane, line 14: I know this is in testing code only, but do we need `fs` here? It would be good not to add on to places we'll need to remove it later.

tests/integrated/test-thumbnail.R
jonkeane: Also, if you put this outside of a `test_that` block, you should be able to just refer to it in each one rather than having to set it each time.

tests/integrated/test-thumbnail.R
toph-allen: Lol, I had to do a `git blame` to figure out what happened here. These were existing `*_image` tests that I had moved to their own file a little while back, but apart from switching them over to use the new functions, I don't think I changed anything, hence the reference to `fs`.

tests/integrated/test-thumbnail.R
toph-allen: Like I replied to the other comment, I haven't futzed with these tests; they are legacy tests adapted to use the new functions.

toph-allen: That to say — you're probably right, and I should check!

tests/integrated/test-thumbnail.R
toph-allen: You can see the block that vanished from `test-deploy.R` to see where they came from.

tests/integrated/test-thumbnail.R
toph-allen: This wasn't a TODO I wrote, this was a legacy TODO (see the left hand side of `tests/integrated/test-deploy.R`).

tests/testthat/test-thumbnail.R
```diff
@@ -1,25 +1,115 @@
...
+    client <- connect(server = "https://connect.example", api_key = "fake")
+    item <- content_item(client, "23456789")
+    expect_error(get_thumbnail(item))
+  })
+})
+
+with_mock_api({
+  test_that("has_thumbnail() returns TRUE when the item has a thumbnail", {
+    client <- connect(server = "https://connect.example", api_key = "fake")
+    item <- content_item(client, "01234567")
```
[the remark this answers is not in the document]
toph-allen: These are technically GUIDs that are redacted, so I'm not sure whether descriptive names would even work, or would be caught either by the redaction code or something internal.

I just found things like this easier to type and recognize than truly random GUIDs obtained from recorded requests.

tests/testthat/test-thumbnail.R
```diff
@@ -1,25 +1,115 @@
...
+  test_that("get_thumbnail() gets the thumbnail", {
+    client <- connect(server = "https://connect.example", api_key = "fake")
+    item <- content_item(client, "01234567")
+    
+    # User-specified path
+    user_path <- tempfile("thumbnail_", fileext = ".jpg")
+    received_path <- get_thumbnail(item, user_path)
+    received <- readBin(received_path, "raw", n = 8)
+    expected <- as.raw(c(0x4e, 0x41))
+    expect_equal(normalizePath(user_path), normalizePath(received_path))
```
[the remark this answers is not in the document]
toph-allen: I definitely added them for a reason, but experimentally, we don't need them any longer.

tests/testthat/test-thumbnail.R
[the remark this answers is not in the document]
toph-allen: I'd love to be wrong and be able to use totally-non-conformant strings for "GUIDs" here.

tests/testthat/setup.R
```diff
@@ -34,5 +34,5 @@ set_redactor(
 
 # Mocks are in directories by Connect version. 2024.08.0 contains all mocks
 # created before 2024.09.0.
-.mockPaths("_mocks/2024.08.0")
+options(httptest.mock.paths = "_mocks/2024.08.0")
```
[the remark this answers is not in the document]
toph-allen: I don't want to have `tests/testthat` on `.mockPaths()`.

R/thumbnail.R
```diff
@@ -85,7 +87,7 @@ has_thumbnail <- function(content) {
   # Connect 2024.09.0 introduced public endpoints for content thumbnails. We
   # prefer those, falling back to the unversioned endpoints if unavailable.
   res <- con$GET(
-    v1_url("content", guid, "thumbnail"),
+    url = con$server_url("content", guid, "__thumbnail__"),
```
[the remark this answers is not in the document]
toph-allen: We can't, because `content/{guid}/__thumbnail__` isn't a `v1` URL.

tests/integrated/test-thumbnail.R
jonkeane: Cool, IMO 🚮 then

tests/testthat/test-thumbnail.R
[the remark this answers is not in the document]
jonkeane: Ah no, you're right. I didn't realize they were guids, my bad. You could look to https://nedbatchelder.com/text/hexwords.html for fun versions of these, but aligning the specific uses/descriptions with one word is too much. 

A comment explaining this guid has these properties, this guid has these other properties might be nice but not necessarily an absolute requirement

toph-allen:
`R CMD check` is failing because we have one non-portable filename:

- `tests/testthat/_mocks/2024.08.0/__api__/v1/oauth/integrations/credentials-fe6213-POST.json`

R/connect.R
```diff
@@ -110,6 +110,12 @@ Connect <- R6::R6Class(
       paste(self$server, "__api__", ..., sep = "/")
     },
 
+    #' @description Build a URL relative to the server root
+    #' @param ... path segments
+    server_url = function(...) {
+      paste(self$server, ..., sep = "/")
+    },
```
jonkeane, line 117: Any reason to use `server_url` in the call to `api_url`? It might be over-engineering if we really only use it for the (old) thumbnails. But in that case, it might be good to mention that here that this is not really something someone should use, and is only here for compatibility (and when that compatibility is removed so too should this)

jonkeane: Oh sorry, actually I have this backwards! The _new_ API has thumbnails at an `__api__`less URL, like `https://connect.company.com/content/25438b83-ea6d-4839-ae8e-53c52ac5f9ce/__thumbnail__` So this helper is needed for those. Are there any other circumstances we would need this? Maybe we should "just" construct that special URL in the thumbnail function rather than exposing here? 

It's definitely confusing to me to have two methods that provide almost the same thing.

R/thumbnail.R
```diff
@@ -0,0 +1,327 @@
...
+    ct <- httr::headers(res)$`content-type`
+    if (grepl("^image/", ct)) {
+      # Just strip off 'image/'
+      ext <- substr(ct, 7, nchar(ct))
+      path <- tempfile(pattern = "content_image_", fileext = paste0(".", ext))
+    } else {
+      # Try png
+      warning(glue::glue("Could not infer file extension from content type: {ct}. Using '.png'"))
+      path <- tempfile(pattern = "content_image_", fileext = ".png")
+    }
```
jonkeane, line 57: Could we do something like `ext <- names(mime::mimemap[mime::mimemap == ct])` using the map directly and so we aren't futzing with a parser here.

Separately / additionally: any reason to assume `png` if we can't figure out what the type is? That seems unhelpful no?

tests/integrated/test-deploy.R
```diff
@@ -269,6 +269,8 @@ test_that("set_image_webshot works", {
     {
       res <- set_image_webshot(cont1_content)
     },
+
+
```
jonkeane, line 273: These are unintended, yeah?

tests/testthat/test-content.R
```diff
@@ -231,4 +231,3 @@ with_mock_api({
     expect_identical(v$key, "WrEKKa77")
   })
 })
-
```
jonkeane, line 234: Errant whitespace change her too, yeah?

R/connect.R
toph-allen: I don't feel strongly either way. I remember feeling at the time like it was better to create a function that lived alongside `api_url` — I think I discussed it with @nealrichardson at the time. I don't remember the specific reason why I went in this direction. T I can definitely imagine other reasons why we might want this, but they're all hypothetical at this point so kinda moot.

tests/testthat/test-thumbnail.R
```diff
@@ -0,0 +1,120 @@
+mock_dirs <- c(
+  "unversioned" = "2024.08.0",
+  "v1" = "2024.09.0"
+)
+
```
jonkeane, line 5: This is a more or less stylistic suggestion, but what if you used the version numbers as the only thing that switched here? As in, don't name them (you can and should still comment and explain that that is where the transition for thumbnails happened, of course). But then you don't need to juggle both names and versions in the test versions down below. 

Thinking about our future selves: when I see the error about `get_thumbnail() gets the thumbnail v1` I'm not sure I would know that I need to go to the top of this file to find the mapping (or that v1 for thumbnails means I'm in the 2024.09.0 directory). But if it were `get_thumbnail() gets the thumbnail 2024.09.0` I have some idea more directly that I should look there.

tests/testthat/test-thumbnail.R
```diff
@@ -0,0 +1,120 @@
...
+      })
+    
+      test_that(glue::glue("has_thumbnail() returns TRUE when the item has a thumbnail ({version})"), {
+        client <- connect(server = "https://connect.example", api_key = "fake")
+        item <- content_item(client, "01234567")
+        expect_true(has_thumbnail(item))
+      })
+    
+      test_that(glue::glue("has_thumbnail() returns FALSE when the status code is 204 ({version})"), {
+        client <- connect(server = "https://connect.example", api_key = "fake")
```
jonkeane, line 64: Can `client <- connect(server = "https://connect.example", api_key = "fake")` be at the top of the `with_mock_api()` block and then reused throughout?

tests/testthat/test-thumbnail.R
```diff
@@ -0,0 +1,120 @@
...
+        expect_false(has_thumbnail(item))
+      })
+    
+      test_that(glue::glue("has_thumbnail() errors with 404 status codes ({version})"), {
+        client <- connect(server = "https://connect.example", api_key = "fake")
+        item <- content_item(client, "23456789")
+        expect_error(has_thumbnail(item), "request failed with Client error: \\(404\\) Not Found")
+      })
+    
+      test_that(glue::glue("set_thumbnail() returns returns a content item when successful ({version})"), {
```
jonkeane, line 75:
```suggestion
      test_that(glue::glue("set_thumbnail() returns returns the content item when successful ({version})"), {
```

Is this more correct? Or maybe I don't understand what's being tested?

tests/testthat/test-thumbnail.R
```diff
@@ -0,0 +1,120 @@
...
+      })
+    
+      test_that(glue::glue("set_thumbnail() works with remote images ({version})"), {
+        client <- connect(server = "https://connect.example", api_key = "fake")
+        item <- content_item(client, "01234567")
+        expect_GET(set_thumbnail(item, "https://other.server/non-connect/working-image/not-an-image"),
+                   "https://other.server/non-connect/working-image/not-an-image")
+        # We're only asserting that this GETs.
+        # We can't use a mock because of this issue:
+        # https://github.com/nealrichardson/httptest/issues/86
```
jonkeane, line 96:
```suggestion
        # We're only asserting that the remote image hits GET. Because we httr::write_disk() and then using what is written doesn't work (https://github.com/nealrichardson/httptest/issues/86)
        # the rest of this function is tested elsewhere, so we have confidence that so long as getting and writing to disk works, we are good.
        expect_GET(set_thumbnail(item, "https://other.server/non-connect/working-image/not-an-image"),
                   "https://other.server/non-connect/working-image/not-an-image")
```

R/connect.R
jonkeane: If that's the case I would move it down and construct the URLs in `*_thumbnail()` with a comment that we can't use the API url constructors cause those URLs are outside that scope, if we run into that more we might consider factoring it out.

But leaving it here with a heavy comment that we question it's needed is something we could do too. Though I suspect it will still confuse us / we won't remember to look here and see that we (probably) should (almost?) never use it

R/thumbnail.R
toph-allen: This code is taken directly from the old implementation. Thinking about the default-PNG assumption — perhaps it's there for a reason, but I'm not sure.

I think it would be reasonable to use `mime::mimemap`. @nealrichardson and I were discussing this earlier (see https://github.com/rstudio/connectapi/pull/303#discussion_r1775910845). The `mimemap` conversion would be basically the same as just stripping the `image/`, except that there would be multiple matches, e.g.:

```
jpeg = "image/jpeg",
jpg = "image/jpeg",
jpe = "image/jpeg",
```

so, `ext <- names(mime::mimemap[mime::mimemap == ct])[1]`.

I don't personally find that much easier to comprehend than `substr(ct, 7, nchar(ct))`. But I guess if someone was using an SVG (is that even possible?) it'd be better to use the map, so yeah, I'm convinced.

tests/testthat/test-content.R
toph-allen: I think there was a double-new-line at the end of this file, and perhaps it was fixed automatically by my editor at some point, or perhaps I fixed it. Either way, the whitespace at the end of the file is correct as it stands.

tests/testthat/test-thumbnail.R
toph-allen: I see your point — how would you feel about having both in the test name?

tests/testthat/test-thumbnail.R
toph-allen: Experimentally: no.

toph-allen: Yes.

tests/testthat/test-thumbnail.R
toph-allen: Yes. Accepted the change locally.

tests/testthat/test-thumbnail.R
toph-allen: Accepting suggestion locally so I can reflow the comment.

tests/testthat/test-thumbnail.R
toph-allen: (I put both in the name)

toph-allen:
Something is broken in the integration tests. I was not running those as I was writing these changes, so who knows…

tests/integrated/test-thumbnail.R
```diff
@@ -0,0 +1,90 @@
+# The tests in this file were not written explicitly for the *_thumbnail()
+# functions; they are taken from `tests/integrated/test-deploy.R` where they
+# test the *_image.R functions.
+
+test_that("set_thumbnail works with local images", {
+  scoped_experimental_silence()
+  img_path <- rprojroot::find_package_root_file("tests/testthat/examples/logo.png")
+
+  res <- set_thumbnail(cont1_content, img_path)
```
jonkeane, line 9: At least some of the integration test issues are because `cont1_content` isn't defined above here (or as far as I can tell anywhere in this file?).

toph-allen: Yeah, I noticed. When I originally made this change, I made some incorrect assumptions about the structure of the integration tests.

toph-allen: Yeah — I didn't copy the preamble from `integrated/test-deploy.R`

R/thumbnail.R
```diff
@@ -0,0 +1,327 @@
...
+
+  if (httr::status_code(res) == 204) {
+    return(NA_character_)
+  }
+
+  # Guess file extension
+  if (is.null(path)) {
+    ct <- httr::headers(res)$`content-type`
+    if (grepl("^image/", ct)) {
+      # Just strip off 'image/'
```
aronatkins, line 50: This comment is inaccurate; this uses mime::mimemap and does not strip an `image/` prefix.

What about using `mime::mimemap` to qualify the content type without looking for an `image/` prefix? Fall back to PNG when it is not a known type.

```r
ext <- if (any(mime::mimemap == ct)) {
    names(mime::mimemap[mime::mimemap == ct])[1]
} else {
    "png"
}
```

R/thumbnail.R
```diff
@@ -0,0 +1,327 @@
...
+      ext <- names(mime::mimemap[mime::mimemap == ct])[1]
+      path <- tempfile(pattern = "content_image_", fileext = paste0(".", ext))
```
aronatkins, line 52: nit: Consider including the GUID in the temporary file name.

R/thumbnail.R
```diff
@@ -0,0 +1,327 @@
...
+  invisible(content)
+}
+
+#' Check content item thumbnail
+#'
+#' Check whether a content item has a thumbnail.
+#'
+#' @param content A content item.
+#' 
+#' @returns `TRUE` if the content item has a thumbnail, otherwise `FALSE`.
```
aronatkins, line 112: note: the true/false are accurate only when the client has permissions to view the thumbnail.

aronatkins (approved):

R/thumbnail.R
toph-allen: Yeah, definitely more elegant.

R/connect.R
nealrichardson: How about:

* Rewrite `api_url()` to be `function(...) self$server_url("__api__", ...)` to make the relationship more clear
* Make sure the use of server_url in the thumbnail function is well commented to call out that it's GETting something that's not under `__api__`

R/thumbnail.R
```diff
@@ -0,0 +1,327 @@
...
+#'
+#' Download the thumbnail for a content item on Connect to a file on your
+#' computer.
+#'
+#' @param content A content item.
+#' @param path Optional. A path to a file used to write the thumbnail image. If
+#' no path is provided, a temporary file with the correct file extension is
+#' created. If the content item does not have a thumbnail, returns `NA`.
+#' 
+#' @returns The path to the image file.
```
nealrichardson, line 11:
```suggestion
#' created.
#' 
#' @returns The path to the downloaded image file, if `content` has a thumbnail; otherwise `NA`.
```

R/thumbnail.R
```diff
@@ -0,0 +1,327 @@
...
+  # Guess file extension
+  if (is.null(path)) {
+    ct <- httr::headers(res)$`content-type`
+    if (grepl("^image/", ct)) {
+      # Just strip off 'image/'
+      ext <- names(mime::mimemap[mime::mimemap == ct])[1]
+      path <- tempfile(pattern = "content_image_", fileext = paste0(".", ext))
+    } else {
+      # Try png
+      warning(glue::glue("Could not infer file extension from content type: {ct}. Using '.png'"))
```
nealrichardson, line 55: Why would we ever be here? Does Connect serve non-images, or not emit Content-Type? And why would assuming png be better than assuming nothing? 

How about this for the logic here?

```
if (is.null(path)) {
  path <- tempfile(pattern = "content_image_")
}

ct <- httr::headers(res)$`content-type`
if (ct %in% mime::mimemap) {
  ext <- names(mime::mimemap[mime::mimemap == ct])[1]
  if (!endsWith(path, ext) {
    path <- paste(path, ext, sep = ".")
  }
}
```

That way, (a) we never assume a wrong type, and (b) we make sure that `path` always ends with the right file extension, even if I provide a mismatched one, so that `httr` will do the right thing with it on upload (since that was the reason you identified for trying to identify the content type in the first place).

toph-allen: Yeah, I was already considering using the determined extension to override the user-provided extension — I like that.

I don't know why the original code uses `png` in the case of content type being indeterminate. Perhaps a better behavior would be to leave the file with no extension and emit a warning; do you like that better?

toph-allen: One wrinkle with this approach verbatim is that you can easily get `tempfile.jpg.jpeg` 😂

tests/testthat/test-thumbnail.R
```diff
@@ -0,0 +1,125 @@
...
+        expect_equal(user_path, received_path)
+        expect_equal(received, expected)
+
+        # User-specified path (wrong extension)
+        user_path <- tempfile("thumbnail_", fileext = ".png")
+        received_path <- get_thumbnail(item, user_path)
+        received <- readBin(received_path, "raw", n = 8)
+        expected <- as.raw(c(0x4e, 0x41))
+        expect_equal(paste0(user_path, ".jpeg"), received_path)
+        expect_equal(received, expected)
```
jonkeane, line 36: This results in `thumbnail_blah.png.jpeg` is that what we want?

.github/workflows/integration-tests.yaml
```diff
@@ -21,7 +21,6 @@ jobs:
           - "2022.09.0" # bionic
           - "2022.03.2"
           - "2021.09.0"
-          - "1.8.8.2"
```
nealrichardson, line 24: Why drop this one?

tests/testthat/test-thumbnail.R
toph-allen: It seems reasonable to me. The converse (substituting rather than appending the correct extension) also seems reasonable.

This only occurs when the user has provided a file path with the incorrect extension. I could see an argument that in this case, it's better to preserve the entire original user-supplied path and just append the correct extension. That's what @nealrichardson did in his suggestion, so I assumed that it was done with that intent.

.github/workflows/integration-tests.yaml
toph-allen: Some of the old integration tests, when applied to the new thumbnail methods, work on all versions listed except 1.8.8.2.

Usually removing a version because tests fail is definitely a no-go, but this is way outside of our support window — I checked [here](https://posit.co/support/#supported-pro-versions) and the oldest supported version is 2023.03. I assumed that its inclusion was an oversight and removed it.

R/connect.R
toph-allen: @nealrichardson I had considered that; I like it. Done.

tests/integrated/test-thumbnail.R
toph-allen: This comment was in relation to old integration tests applied verbatim to new thumbnail functions. The integration tests need a more thorough look in general outside of this PR. Jon and I discussed this in an off-GitHub conversation at some point. I'm resolving this conversation to indicate that.

.github/workflows/integration-tests.yaml
nealrichardson: Can we just skip those tests on older Connect versions? We have a test helper for that IIRC. 

I know this is outside of our support window, but if it's not a burden to us (parallel CI is amazing), I'd like to keep as broad support as is reasonable. We do have many customers running Connect beyond the support window, and our SDKs can be a useful tool to help them manage and upgrade their content to help with upgrading Connect itself.

tests/testthat/test-thumbnail.R
nealrichardson: Right. IMO I don't think it's worth trying to be even more clever with the filenames.

.github/workflows/integration-tests.yaml
toph-allen: Sure thing! Keeping support as broad as possible makes sense, especially given that the SDKs are helpful when people need to migrate to newer, supported versions. I'll look into skipping those tests on 1.8.8.2.

toph-allen: The failing test is skipped now (technically two tests, but both wind up calling `set_thumbnail()` with a remote path to Connect's favicon). Much better than omitting _all_ tests on that version. :)

PR #315, opened 2024-10-11
chore: use v1 timezones API

### LINKED ISSUE #300: use /__api__/v1/timezones rather than /__api__/timezones (2024.09.0)
The Connect 2024.09.0 release publicizes the time zone enumeration API.

`connectapi::get_timezones()` needs to call this new endpoint when available and call the unversioned endpoint otherwise.

https://github.com/rstudio/connectapi/blob/bfefb00ce4721c0bb80aedf58ae217ff63742bee/R/schedule.R#L339-L345

The payload is unchanged; only the request path is different.

toph-allen:
## Intent

This PR contains two sets of changes, which were interrelated at the start of this PR but are now less so:

- Use v1 timezones API when available.
- Refactor some server-version-related code.
- A different approach to mocking HTTP methods.

Fixes #300

### Time zones

The time zones change is relatively simple.

The server-version-related refactoring is in here because I originally was using the time zones API switch to think about how the logic for choosing a URL would look if it also keyed off the server version. It made the logic much more complicated than was necessary, so I removed that logic from the `get_timezones()` function, but I have kept some of the version-getting-related refactoring. Another option would be moving it to a different PR.

### Server version

The gist of the refactoring is:

- Renaming some private functions.
- `Connect$version` is now an active binding on the `Connect` object, a private property that is `NULL` by default, and a character vector when fetched. `NA` indicates that the version is hidden.

This could easily be removed from this PR and tabled or submitted in a separate PR.

### Testing

I've been finding it difficult to manage our testing fixtures here — we have 79 files in our fixtures dirs right now, and managing tests when the required context is distributed in that file hierarchy has been a cognitive challenge. For simple unit tests like this, I thought an approach where I stub out the `request` method in the subclass would provide a few benefits.

- Test a variety of different responses from the same endpoint without needing different mock dirs.
- Track the exact requests that were made of the object and confirm expected behavior there.

## Checklist


R/schedule.R
```diff
@@ -337,7 +337,26 @@ schedule_describe <- function(.schedule) {
...
+    if (httr::status_code(res) == 404) {
+      # We will try the old URL.
+      res <- NULL 
+    }
+  }
+  if (is.null(res)) {
+    res <- connect$GET(unversioned_url("timezones"), parser = NULL)
+    print(res)
+  }
+  connect$raise_error(res)
```
jonkeane, line 356: It feels odd to me that we both check the version and then also later check for a 404 and `NULL` the result. What if instead we did something like:

```
res <- tryCatch(
  connect$GET(v1_url("timezones"), parser = NULL),
  error = function(e) {
    res <- connect$GET(unversioned_url("timezones"), parser = NULL)
  }
)
connect$raise_error(res)
```

Would that work? If no, what would we need to make that work?

toph-allen: I don't _think_ this quite works.

(1) We had discussed using cached version information on the server to avoid calling the new endpoint if we _know_ Connect is of an older version than has the new endpoint. When we do that, we still need to be able to fall back to just calling both endpoints if the version is unavailable (`Server.HideVersion`). The timezones API isn't one where the cost of two requests is significant, but it's really simple, so I was using it to experiment with that logic.

(2) `connect$GET` only throws an error if `parser` is not `NULL` (see [here](https://github.com/rstudio/connectapi/blob/7762c7acef13145864518482d636278542782ef5/R/connect.R#L126-L148)). This means that we'd never get to the `error` condition in that `tryCatch`. I think the most concise possible version of this logic that does not check the Connect version is in the thumbnail functions (example [here](https://github.com/rstudio/connectapi/blob/7762c7acef13145864518482d636278542782ef5/R/thumbnail.R#L31-L42)).

Adapting that here would look like:

```r
res <- con$GET(
  url = con$v1("timezones"),
  parser = NULL
)
if (httr::status_code(res) == 404) {
  res <- con$GET(
    unversioned_url("timezones"),
    parser = NULL
  )
}
con$raise_error(res)
```

That doesn't check the Connect version, though, and it'd still be nice to avoid calling v1 URLs if we _know_ they aren't available. But perhaps that's a low enough priority that it's not worth implementing here. Or perhaps there's a better way to do it, e.g. registering implementations based on Connect's version when the client is created, which would be totally outside of the scope of this PR.

toph-allen: Yeah, I get that. Since we had discussed this approach, I figured a simpler API was a good moment to experiment with it — but I'll cherry-pick those changes into a separate branch and revert/simplify them here.

I think a `tryCatch()` approach can probably work, if we don't pass in `parser = NULL`, because that's what suppresses raising errors (which happens in `Connect$request()`).

tests/testthat/test-schedule.R
```diff
@@ -0,0 +1,211 @@
...
+}
+
+
+mock_response(
+  url = "v1/timezones",
+  body = list(
+    list(timezone = "Africa/Abidjan", offset = "+00:00"),
+    list(timezone = "Africa/Accra", offset = "+00:00")
+  )
+)
```
jonkeane, line 78: Is this something you're having trouble getting `httptest` to do? This pattern should be possible: 

https://github.com/Crunch-io/rcrunch/blob/b709e282172ee6b813b26299c28514ea840c47a6/tests/testthat/test-api.R#L117-L124 and the fixture https://github.com/Crunch-io/rcrunch/blob/b709e282172ee6b813b26299c28514ea840c47a6/tests/testthat/app.crunch.io/404.R#L1-L18 are one example of this.

I know that you mentioned wanting to experiment with possible other test runners / setups for the http testing, but I don't think this is the place we should do that. Let's keep using what we have and defer taking a new testing approach until after we have at least a few PRs on our journey of improving the recipes.

toph-allen: This commit was pushed up in a state I had not intended to receive review. I had a 1-on-1 with @nealrichardson  earlier today and asked some questions about the approach to testing that I had taken with this PR, and he gave some thoughts on streamlining it that I implemented after our chat.

I was motivated to explore this approach as I was adding fixtures for the tests yesterday. There are three scenarios I want to test:

1. Both endpoints respond
2. The new endpoint returns a 404
3. The new and old endpoints returns a 404

So far, we have two mock dirs that are used for different versions of Connect. I can use those to mark out the two scenarios, but to test a third scenario requires adding another mock directory, and — I'm loathe to move to a model where we have additional mock dirs for each testing scenario, and managing tests of simple scenarios that are split out across an increasingly complex file tree has been a headache.

I threw this together in a few hours. It seems to work, and I @nealrichardson's suggestions made it much tighter. I also have found the workflow a good fit for me personally, and I think the tests are nicely concise and contained — it feels like a good fit for this kind of narrowly-scoped unit test.

I'm actually really excited about this approach — I feel like I've solved a piece of workflow friction that has been slowing me down.

tests/testthat/test-schedule.R
```diff
@@ -0,0 +1,87 @@
...
+
+test_that("get_timezones() gets timeszones from unversioned url when v1 returns 404", {
+  client <- MockConnect$new()
+
+  client$mock_response(
+    path = "v1/timezones",
+    content = "404 page not found",
+    status_code = 404L,
+    headers = c("Content-Type" = "text/plain; charset=utf-8")
+  )
```
tdstein, line 38: I personally like this abstraction quite a bit. It removes a lot of the boilerplate and clarifies what's being augmented for the purpose of the test. 

Also, This looks fairly similar to the `responses` package we are using in Python.

tdstein (approved):
Implementation looks good to me. Have we had a chance to test this against live versions of the relevant Connect versions?

toph-allen:
I've tested it against Dogfood. But -- what I also am reminded is that I should try to add an integration test for this.

toph-allen:
The `get_timezones()` function failed in the integration tests. The test was previously commented out, and I tried a different approach, but I suspect there might be something about GitHub Actions runners which make it fail.

I manually built Connect 2024.08.0 locally and tested against that to manually verify against relevant versions.

tests/testthat/test-schedule.R
```diff
@@ -0,0 +1,87 @@
...
+      list(timezone = "Africa/Accra", offset = "+00:00")
+    )
+  )
+  client$mock_response(
+    path = "timezones",
+    content = list(
+      list(timezone = "Africa/Abidjan", offset = "+00:00"),
+      list(timezone = "Africa/Accra", offset = "+00:00")
+    )
+  )
```
nealrichardson, line 18: If you're calling this at all, there's a problem, so no need for it to exist.

```suggestion
```

tests/testthat/setup.R
```diff
@@ -35,3 +36,67 @@ set_redactor(
...
+
+      if (is.null(parser)) {
+        res
+      } else {
+        self$raise_error(res)
+        httr::content(res, as = parser)
+      }
+    },
+    responses = list(),
+    mock_response = function(path, content, status_code = 200L, headers = c("Content-Type" = "application/json; charset=utf-8")) {
```
nealrichardson, line 67: Obviously you're only doing GETs here so you don't hit this, but it's odd that there's no request method or other stuff here. The assurance that you're making the correct request isn't great, though at least you're checking the request method in the call log assertions.

tests/integrated/test-schedule.R
```diff
@@ -80,9 +80,10 @@ test_that("schedule display works", {
   expect_snapshot_output(schedule_describe(tmp))
 })
 
+# Doesn't work in integration tests for some reason.
```
nealrichardson, line 83: Then delete this? No good leaving commented code around. 

This should be solvable I would think, set something in the Docker image and assert we're picking it up here. IDK that it's worth the effort for such an insignificant API though.

tests/testthat/setup.R
toph-allen: Yeah — I'll want to add a verb argument. The approach of constructing the response list names from `"{verb} {url}"` would work for now — so, looking requests up by the string that's recorded in the `call_log` field.

tests/integrated/test-schedule.R
toph-allen: Will do.

PR #318, opened 2024-10-28
feat: add functions to `get_runtime_caches()` and `delete_runtime_caches()`

### LINKED ISSUE #312: Support cookbook in `connectapi`: Viewing Runtime Caches, Deleting Runtime Caches
- [Viewing Runtime Caches](https://docs.posit.co/connect/cookbook/server-management/viewing-runtime-caches/index.html)
- [Deleting Runtime Caches](https://docs.posit.co/connect/cookbook/server-management/deleting-runtime-caches/index.html)

### Viewing Runtime Caches

```r
library(connectapi)

client <- connect()
get_runtime_caches(client)
```

```r
# A tibble: 23 × 3
   language version image_name
   <chr>    <chr>   <chr>
 1 R        3.0.2   Local
 2 R        3.1.0   Local
 3 R        3.1.2   Local
 4 R        3.1.3   Local
 5 R        3.2.0   Local
 6 R        3.2.1   Local
 7 R        3.2.2   Local
 8 R        3.3.3   Local
 9 R        3.4.0   Local
10 R        3.4.4   Local
# ℹ 13 more rows
# ℹ Use `print(n = ...)` to see more rows
```

Notes:

- Perhaps this means that #311's function should be something like `get_runtime_versions()`. 🤔 
- Worth implementing filtering by language?

### Deleting Runtime Caches

```r
library(connectapi)

client <- connect()
delete_runtime_cache(language = "r", version = "3.0.2")
```

```
Deleting runtime cache...
Successfully deleted runtime cache
```

Notes

- `image_name` should default to `"Local"` so is only required for off-host execution.
- The server returns a task object. Perhaps it makes sense to have this function poll the task and only return when deletion is complete.

toph-allen:
## Intent

Add support for getting and deleting runtime caches.

- `get_runtime_caches(client)` returns a tibble of runtime caches.
- `delete_runtime_cache(client, language, version, image_name, dry_run)` deletes a cache.

```r
> client <- connect()
Defining Connect with server: connect.example)
> get_runtime_caches(client)
# A tibble: 22 × 3
   language version image_name
   <chr>    <chr>   <chr>     
 1 R        3.1.0   Local     
 2 R        3.1.2   Local     
 3 R        3.1.3   Local     
 4 R        3.2.0   Local     
 5 R        3.2.1   Local     
 6 R        3.2.2   Local     
 7 R        3.3.3   Local     
 8 R        3.4.0   Local     
 9 R        3.4.4   Local     
10 R        3.5.3   Local     
# ℹ 12 more rows
# ℹ Use `print(n = ...)` to see more rows
```

Fixes #312

## Notes

- Nothing fancy.
- Tests use `httptest` mocks here.
- Fixed a bug where `Connect$DELETE()` didn't respect the `parser` argument.
- I think there might be a bug in our error handling, because the error I get when the response 404s happens _in_ `Connect$raise_error()` (see below)

From vscode so may be formatted weird:

```r
> client$raise_error(res)
List of 3 $ code   : int 4 $ error  : chr "Cache does not exist" $ payload: NULL
Error in `client$raise_error()`:
! https://rsc.radixu.com/__api__/v1/system/caches/runtime request failed with Client error: (404) Not Found
Show Traceback
```

## Checklist


R/runtime-caches.R
```diff
@@ -0,0 +1,73 @@
...
+  purrr::map_dfr(res$caches, ~.x)
+}
+
+#' Delete a runtime cache
+#' 
+#' Delete a runtime cache from a Connect server.
+#' 
+#' @param client A `Connect` object.
+#' @param language The language of the cache, either "R" or "Python".
+#' @param version The version of the cache, e.g. "3.6.3".
```
schloerke, line 31: Let's update to a recent version

```suggestion
#' @param version The version of the cache, e.g. "4.3.3".
```

R/runtime-caches.R
```diff
@@ -0,0 +1,73 @@
...
+#'   the deletion.
+#' 
+#'
+#' @returns A `Task` object representing the deletion task. If `dry_run` is
+#' `TRUE`, returns `NULL` or throws an error if the deletion would fail.
+#'
+#' @examples
+#' \dontrun{
+#' client <- connect()
+#' task <- delete_runtime_cache(client, "R", "3.6.3")
```
schloerke, line 44:
```suggestion
#' task <- delete_runtime_cache(client, "R", "4.3.3")
```

R/runtime-caches.R
```diff
@@ -0,0 +1,73 @@
...
+    ),
+    encode = "json",
+    parser = "parsed"
+  )
+
+  if (dry_run == TRUE) {
+    message(glue::glue(
+      "Runtime cache deletion dry run finished; ",
+      "language = {language}, version = {version}, image_name = {image_name}"
+    ))
```
schloerke, line 67: I don't know how other similar parts of the code work...

Is there anything from `res` that can be included here?

schloerke (approved):
(Given small updates for comments)

toph-allen:
I made a change that fixed the `pkgdown` failure. But I'm confused as to why it only complained about `delete_runtime_cache()` and not `get_runtime_caches()` — it wants the former in the index but does _not_ want the latter in it. Is there a mistake in my `roxygen2` comments causing that discrepancy? I didn't set up the `pkgdown` site so I don't know if there are some settings I'm unaware of causing this.

R/runtime-caches.R
toph-allen: There isn't really anything meaningful in the response — it's basically a `200` if the directory exists and has the requisite permissions to delete (I think), and an error if not (and we report the error). The success is pretty information-free. Happy to take suggestions for better wording though. This wording is taken from the `rsconnect-python` implementation.

R/runtime-caches.R
```diff
@@ -0,0 +1,72 @@
...
+#' 
+#' @family server management functions
+#' @export
+get_runtime_caches <- function(client) {
+  validate_R6_class(client, "Connect")
+  res <- client$GET(v1_url("system/caches/runtime"))
+  purrr::map_dfr(res$caches, ~.x)
+}
+
+#' Delete a runtime cache
```
tdstein, line 25: Do we need to include any information regarding permissions in the documentation?

tdstein (approved):
I have a question regarding documentation. Otherwise, the implementation looks good to go!

R/runtime-caches.R
toph-allen: Good point, I'll add a note to that effect.

R/runtime-caches.R
toph-allen: I think I'll change it to say "did not encounter any errors", as that's kind of a better representation of what it means.

nealrichardson (commented):
Sigh, this was still a "pending" review from the other day. Sorry this is late, hope there's something useful still in there.

toph-allen:
Still helpful -- I'm mostly OOO for the rest of the week but I'll push up some tweaks in response to your comments.

PR #323, opened 2024-11-07
chore: Remove unused function

schloerke:
While removing manual `nolint` comments, I ran into this function. 

I can not find the function name anywhere in the repo. So it feels fair to remove it.


toph-allen (approved):
I wondered about this function when I added the `nolint` comment! But just assumed that it was used. Thanks for digging deeper.

PR #326, opened 2024-11-07
fix: prevent hanging when calling `get_groups` with `prefix` and small `page_size`

### LINKED ISSUE #319: Investigate error with pagination
- The error is mentioned in a comment in `groups_create_remote()`.

When the page size and number of results would require multiple pages, we instead see repeat pages. Is this a Connect error or a `connectapi` bug?

```r
> get_groups(client, page_size = 2, prefix = "c", limit = 10)
# A tibble: 10 × 3
   guid                                 name                 owner_guid                          
   <chr>                                <chr>                <chr>                               
 1 a6fb7cea- connect_dev          1a775703-
 2 68bad75b- cool_kids_of_the_dmv 1270410b-
 3 a6fb5cea- connect_dev          1a7a5703-
 4 68bad75b- cool_kids_of_the_dmv 1270410b-
 5 a6fb5cea- connect_dev          1a7a5703-
 6 68bad75b- cool_kids_of_the_dmv 1270410b-
 7 a6fb5cea- connect_dev          1a7a5703-
 8 68bad75b- cool_kids_of_the_dmv 1270410b-
 9 a6fb5cea- connect_dev          1a7a5703-
10 68bad75b- cool_kids_of_the_dmv 1270410b-
```

Originally committed here: https://github.com/rstudio/connectapi/commit/bec0b44e3f3513fa85edcf31d3bda0fbb27d23c2

toph-allen:
## Intent

`get_groups(client, prefix = "prefix")` would fail when the `page_size` required pagination — it would hang forever.

Fixes #319

## Approach

- This hang occur because the `v1/groups` API only ever returns the first page of results when using `prefix`.
- Whereas `v1/users` returns no results when a subsequent page of a `prefix` request is called, `v1/groups` always returns the first page, which causes `connectapi` to just page eternally.
- So, we just don't call the pagination function when we're using a prefix!
- Updated documentation to clarify behavior of different parameters.

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,8 +34,7 @@
   available. (#300)
 - `connect$DELETE()` now respects the `parser` argument rather than always using
   `NULL`.
-- `get_groups()` does not attempt to paginate API requests using a search
-  prefix, as this would lead to the function hanging. (#319)
+- `get_groups()` no longer hangs when a search `prefix` is provided. (#319)
 
 # connectapi 0.3.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,6 +34,8 @@
   available. (#300)
 - `connect$DELETE()` now respects the `parser` argument rather than always using
   `NULL`.
+- `get_groups()` does not attempt to paginate API requests using a search
+  prefix, as this would lead to the function hanging. (#319)
 
 # connectapi 0.3.0
 
--- a/R/get.R
+++ b/R/get.R
@@ -60,7 +60,7 @@ get_users <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
 #' @param prefix Filters groups by prefix (group name).
 #' The filter is case insensitive.
 #' @param limit The number of groups to retrieve before paging stops.
-#' 
+#'
 #' `limit` will be ignored is `prefix` is not `NULL`.
 #' To limit results when `prefix` is not `NULL`, change `page_size`.
 #'
--- a/R/get.R
+++ b/R/get.R
@@ -89,11 +89,14 @@ get_users <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
 get_groups <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
   validate_R6_class(src, "Connect")
 
+  # The `v1/groups` endpoint always returns the first page when `prefix` is
+  # specified, so the page_offset function, which increments until it hits an
+  # empty page, fails.
   if (!is.null(prefix)) {
     response <- src$groups(page_size = page_size, prefix = prefix)
     res <- response$results
   } else {
-    res <- page_offset(src, src$groups(page_size = page_size, prefix = prefix), limit = limit)
+    res <- page_offset(src, src$groups(page_size = page_size, prefix = NULL), limit = limit)
   }
 
   out <- parse_connectapi_typed(res, connectapi_ptypes$groups)
--- a/R/get.R
+++ b/R/get.R
@@ -55,11 +55,14 @@ get_users <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
 
 #' Get group information from the Posit Connect server
 #'
-#' @param src The source object
-#' @param page_size the number of records to return per page (max 500)
+#' @param src The source object.
+#' @param page_size The number of records to return per page (max 500).
 #' @param prefix Filters groups by prefix (group name).
 #' The filter is case insensitive.
-#' @param limit The max number of groups to return.
+#' @param limit The number of groups to retrieve before paging stops.
+#' 
+#' `limit` will be ignored is `prefix` is not `NULL`.
+#' To limit results when `prefix` is not `NULL`, change `page_size`.
 #'
 #' @return
 #' A tibble with the following columns:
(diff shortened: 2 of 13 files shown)


tdstein:
> This hang occur because the v1/groups API only ever returns the first page of results when using prefix.

Is this the rumored bug in Connect?

tdstein (approved):
Are there any unit tests that already verify this functionality?

NEWS.md
```diff
@@ -34,6 +34,8 @@
   available. (#300)
 - `connect$DELETE()` now respects the `parser` argument rather than always using
   `NULL`.
+- `get_groups()` does not attempt to paginate API requests using a search
+  prefix, as this would lead to the function hanging. (#319)
```
nealrichardson, line 38: Maybe a little less technical of a note here, like this? (Assuming it is correct)

```suggestion
- `get_groups()` no longer hangs when a search `prefix` is provided. (#319)
```

R/get.R
```diff
@@ -86,7 +89,12 @@ get_users <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
 get_groups <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
   validate_R6_class(src, "Connect")
 
-  res <- page_offset(src, src$groups(page_size = page_size, prefix = prefix), limit = limit)
+  if (!is.null(prefix)) {
```
nealrichardson, line 92: Since this is clearly working around a bug, can we make a followup issue to remove the workaround when the bug is fixed, and link it here? At a minimum, this deserves a comment as to why we're doing this nonstandard thing.

R/get.R
```diff
@@ -86,7 +89,12 @@ get_users <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
...
+    response <- src$groups(page_size = page_size, prefix = prefix)
+    res <- response$results
+  } else {
+    res <- page_offset(src, src$groups(page_size = page_size, prefix = prefix), limit = limit)
```
jonkeane, line 96: `prefix` here will always be `NULL` correct? Would it make sense to do `res <- page_offset(src, src$groups(page_size = page_size), limit = limit)` instead or possibly `res <- page_offset(src, src$groups(page_size = page_size, prefix = NULL), limit = limit)` if you're worried that the default of the `groups` method might change under you? (Though that's also exactly where the tests that Taylor asked about would help of course!)

R/get.R
```diff
@@ -86,7 +89,12 @@ get_users <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
 get_groups <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
   validate_R6_class(src, "Connect")
 
-  res <- page_offset(src, src$groups(page_size = page_size, prefix = prefix), limit = limit)
+  if (!is.null(prefix)) {
+    response <- src$groups(page_size = page_size, prefix = prefix)
```
nealrichardson, line 93: If we don't want pagination in this case, why are we passing a `page_size` argument through here? (And to pull that thread more: why does this function (method) take a page_size argument if it doesn't do pagination?? And then, why does it exist at all?)

toph-allen: When using prefix, the API only ever provides the first page of results, but the number of results on that page are still governed by `page_size`. We edited the documentation of `get_groups` to try to describe this distinction.

R/get.R
toph-allen: Yeah, either of those (and ensuring that this is covered by tests) would make sense.

R/get.R
toph-allen: I'll add a comment to the code explaining why we're doing the non-standard thing. I don't think it's actually a *bug* in Connect (see the response I'm about to write in the main comments thread on this issue in response to @tdstein)

toph-allen:
> > This hang occur because the v1/groups API only ever returns the first page of results when using prefix.
> 
> Is this the rumored bug in Connect?

@tdstein I mean, it's an odd behavior, but the `v1/groups` endpoint works exactly as described in its [documentation](https://docs.posit.co/connect/api/#get-/v1/groups):

> - For a `prefix` search, results are sorted based on similarity to the `prefix`. A `prefix` search ignores `asc_order`. The first page of results will always be returned.

So when calling `v1/groups?prefix=c&page_size=2&page_number=2`, the `results` returned are the same as for page 1. This behavior is kinda weird.

Aside: the `v1/users` endpoint [docs](https://docs.posit.co/connect/api/#get-/v1/users) say that it also behaves this way, but it does not. When calling it with a `prefix` with `page_number` > 1, _no_ results are returned (which I feel like is the most sensible behavior.

I found Cole's original issue about this: https://github.com/rstudio/connect/issues/17543. Two notes on that: it (1) requests that the API return an error with `prefix` and `page_number` > 1, and (2) points to our docs on offset pagination, which I think point to a sensible answer.

The Cookbook docs on offset pagination are [here](https://docs.posit.co/connect/cookbook/pagination/#offset-pagination). They recommend paging through results until you reach an empty page, which suggests that the behavior of the `v1/users` endpoint is the better approach and that the behavior of `v1/groups` endpoint is actually wrong.

Another option for _this_ PR is to have the `page_offset` function check to see if the current result is zero or the same as the previous result, which would make it robust to current and future variations of Connect's API behavior.

toph-allen:
Two Connect bugs:

- https://github.com/rstudio/connect/issues/28774
- https://github.com/rstudio/connect/issues/28773

PR #335, opened 2024-11-19
feat: filter users by `account_status`, `user_role`

### LINKED ISSUE #331: feat: support finding users by `user_role` and `account_status`
https://github.com/rstudio/connect/issues/28852

Add support for `user_role` and `account_status` parameters to `get_users()` and `Connect$users()`.

toph-allen:
## Intent

Add support for filter parameters in `get_users()`: `account_status`, `user_role`

Fixes #331

## Approach

Very simple, just add the parameters. You can pass them as R vectors and they're bundled up into the `foo|bar` Connect uses for boolean OR operations.

## Checklist

### DIFF
--- a/R/get.R
+++ b/R/get.R
@@ -5,6 +5,10 @@
 #' @param prefix Filters users by prefix (username, first name, or last name).
 #' The filter is case insensitive.
 #' @param limit The max number of records to return
+#' @param user_role Filter by user role ("administrator", "publisher",
+#' "viewer"). Pass in a vector of multiple roles to match any value.
+#' @param account_status Filter by account status ("locked", "licensed",
+#' "inactive"). Pass a vector of multiple statuses to match any value.
 #'
 #' @return
 #' A tibble with the following columns:
@@ -34,17 +38,36 @@
 #' library(connectapi)
 #' client <- connect()
 #'
-#' # get all users
-#' get_users(client, limit = Inf)
+#' # Get all users
+#' get_users(client)
+#'
+#' # Get all licensed users
+#' get_users(client, account_status = "licensed")
+#'
+#' # Get all users who are administrators or publishers
+#' get_users(client, user_role = c("administrator", "publisher"))
 #' }
 #'
 #' @export
-get_users <- function(src, page_size = 500, prefix = NULL, limit = Inf) {
+get_users <- function(
+  src,
+  page_size = 500,
+  prefix = NULL,
+  limit = Inf,
+  page_number = NULL,
+  user_role = NULL,
+  account_status = NULL
+) {
   validate_R6_class(src, "Connect")
 
   res <- page_offset(
     src,
-    src$users(page_size = page_size, prefix = prefix),
+    src$users(
+      page_size = page_size,
+      prefix = prefix,
+      user_role = user_role,
+      account_status = account_status
+    ),
     limit = limit
   )
 
--- a/R/connect.R
+++ b/R/connect.R
@@ -506,12 +506,28 @@ Connect <- R6::R6Class(
     #' @param page_number The page number.
     #' @param prefix The search term.
     #' @param page_size The page size.
-    users = function(page_number = 1, prefix = NULL, page_size = 500) {
+    #' @param user_role Filter by user role.
+    #' @param account_status Filter by account status.
+    users = function(
+      page_number = 1,
+      prefix = NULL,
+      page_size = 500,
+      user_role = NULL,
+      account_status = NULL
+    ) {
       path <- v1_url("users")
+      if (!is.null(user_role)) {
+        user_role <- paste(user_role, collapse = "|")
+      }
+      if (!is.null(account_status)) {
+        account_status <- paste(account_status, collapse = "|")
+      }
       query <- list(
         page_number = page_number,
         page_size = valid_page_size(page_size),
-        prefix = prefix
+        prefix = prefix,
+        user_role = user_role,
+        account_status = account_status
       )
       self$GET(path, query = query)
     },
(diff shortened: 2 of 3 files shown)


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Type: Package
 Package: connectapi
 Title: Utilities for Interacting with the 'Posit Connect' Server API
-Version: 0.4.0
+Version: 0.4.0.9000
```
jonkeane, line 4: Do we have on our checklist(s) for release to bump this after the release? I thought it was on the `usethis` one, but if not we should do that so we don't have to tack these onto PRs like this

.Rbuildignore
```diff
@@ -2,6 +2,9 @@
 ^\.Rproj\.user$
 ^\.Renviron$
 ^\.github$
+^\.envrc$
+^\.vscode$
+^.*\.lic$
```
jonkeane, line 7: I don't mind adding these, but IMO they should go separately (like as part of the version bump) rather than a feature PR. I'm fine to keep them here if that's quickest, but it's a funk we should avoid if possible

DESCRIPTION
toph-allen: Agreed. It isn't on our checklist, and I don't *think* it's on the `usethis` one either.

toph-allen: There are a few outdated things on the `usethis` issue template that don't fit so I was thinking of just making an issue template for release prep or something.

R/get.R
```diff
@@ -5,6 +5,10 @@
 #' @param prefix Filters users by prefix (username, first name, or last name).
 #' The filter is case insensitive.
 #' @param limit The max number of records to return
+#' @param user_role Filter by user role ("administrator", "publisher",
+#' "viewer"). Pass in a vector of multiple roles to match any value.
+#' @param account_status Filter by account status ("locked", "licensed",
+#' "inactive"). Pass a vector of multiple statuses to match any value.
```
jonkeane, line 11: Should we tell folks we default to `NULL` and that that means ...?

jonkeane (changes requested):
Thanks for this! 

Would you mind adding a unit test or two? The logic of collapsing isn't massive, but it would still be good practice adding unit tests like those.

toph-allen:
Yep, I'm gonna add unit tests 🫡

.Rbuildignore
toph-allen: Good point -- I'll move them to another PR.

DESCRIPTION
jonkeane: Could also send a PR to usethis too if it's general?

toph-allen: Yeah. I had been considering that for other items -- like there are some list items that refer to packages that are (as far as I can tell) no longer available.

tests/testthat/test-get.R
```diff
@@ -140,3 +140,33 @@ with_mock_api({
...
+    expect_GET(
+      get_users(client),
+      "https://connect.example/__api__/v1/users?page_number=1&page_size=500"
+    )
+
+    # Filter just on one parameter
+    expect_GET(
+      get_users(client, user_role = "administrator"),
+      "https://connect.example/__api__/v1/users?page_number=1&page_size=500&user_role=administrator"
+    )
```
jonkeane, line 157: 🎉

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # connectapi (development version)
 
+## New features
+
+- `get_users()` now supports filtering users with the `account_status` and
+  `user_role` parameters. This allows you to, for example, find all licensed
+  users on a Connect server. (#311)
```
jonkeane, line 7:
```suggestion
- `get_users()` now supports filtering users with the `account_status` and
  `user_role` parameters. For example, this allows you to find all licensed
  users on a Connect server. (#311)
```

jonkeane: Super minor tweak, but trying to make the example bit there to not break up the rest of the sentence

R/get.R
```diff
@@ -5,6 +5,13 @@
 #' @param prefix Filters users by prefix (username, first name, or last name).
 #' The filter is case insensitive.
 #' @param limit The max number of records to return
+#' @param user_role Optionally filter by user role ("administrator",
+#' "publisher", "viewer"). Pass in a vector of multiple roles to match any value
+#' (boolean OR). When `NULL` (the default), results are not filtered.
```
jonkeane, line 10:
```suggestion
#' (boolean OR). When `NULL` (the default), all results are returned.
```

Feel free to take or leave this suggestion, but is this true? That not filtered == all results? I believe yes for sure for `user_role`, but not sure about `account_status` are locked/inactive accounts included if you don't specify anything?

As for the suggestion of "all results" vs "results not filtered" either is fine, the first seems slightly more approachable to me, but I don't super strongly hold that.

jonkeane (approved):
A few minor comments, but nothing that would stop merging.

R/get.R
toph-allen: Yep, this is true! The default for `account_status` is to not filter by status when nothing's passed in — this was my understanding from reading the API docs, and I just confirmed it experimentally (by filtering for all types, which uses boolean OR, and comparing that to unfiltered results).

I agree that the current wording isn't a paragon of accessibility, but to my ear the suggested change sounds almost tautological, in the sense that "result" reads to me as meaning "the set of accounts returned by the function/API call. I wonder if this sentence is even needed — to me, the "Optionally" at the start of the parameter definition says everything that this sentence is trying to say (i.e. this is an optional parameter, and if you don't pass anything, the filtering operation described doesn't happen).

jonkeane: Yeah, I can see arguments for both ways. I do think we should mention the default, it's not good to not say what the default is. Like I said, feel free to reject the suggestion — maybe leaning heavily on `filter` and what that means is enough here.

PR #337, opened 2024-11-21
feat: `get_group_content()` lets you get content that groups can access [draft, tests forthcoming]

### LINKED ISSUE #334: feat: support getting the content a group has access to

toph-allen:
## Intent

Add a function to support getting content that a group has access to.

Fixes #334 

## Approach

The new function, `get_group_content()`, requires a `src` argument and `groups`, which can either be a data frame with `guid` and `name` columns or a character vector of GUIDs.

```r
client <- connect()
groups <- get_groups(client)

# one group
groups %>%
  filter(name == "research") %>%
  get_group_content(client, .)

# all groups
get_group_content(client, groups)

# use the GUID only
get_group_content(client, groups$guid)
```

This is slightly different from similar functions like `get_group_members(client, group_guid)`, which requires only the guid. Reasons for the difference:

- Just passing in the `group_guid` is marginally simpler for the user for getting one group's content (just select the row, don't need to pull the column), and significantly simpler for getting multiple groups (just pass in a data frame with multiple rows). Landed upon this approach & reasoning rubber-ducking with @schloerke earlier last week.
- It allows us to include `group_name` in the returned data frame with less overhead. The downside of this is, to keep our return shape consistent, we need to look up the `group_name` for each group if passed a bare character vector of GUIDs.

## Other notes

- I moved all groups-related functions to a `groups.R` file.
- Added a utility function to verify the schema of the input data frame. It seemed like the function was getting complex enough to factor it out anyway, and this pattern seems like a very useful one given that so many of our functions return data frames.

## Checklist


R/groups.R
```diff
@@ -0,0 +1,184 @@
...
+#' dplyr::filter(groups, name = "research_scientists") %>%
+#'   get_group_content(client, groups = .)
+#'
+#' # Get permissions for all groups by passing in the entire groups data frame.
+#' get_group_content(client, groups)
+#' }
+#'
+#' @family groups functions
+#' @export
+get_group_content <- function(src, groups) {
```
schloerke, line 144: With `groups` being a _newish_ parameter type, Can we add a small "this is a data frame and has `guid` and `name` columns?" check?

R/groups.R
```diff
@@ -0,0 +1,184 @@
...
+get_group_content_impl <- function(src, guid, name) {
+  validate_R6_class(src, "Connect")
+
+  res <- src$group_content(guid)
+  parsed <- parse_connectapi_typed(res, connectapi_ptypes$group_content)
+
+  dplyr::transmute(parsed,
+    group_guid = guid,
+    group_name = name,
+    content_guid,
```
schloerke, line 163: These lintr warnings (and R CMD check warnings) should be able to be removed if `.data$content_guid` is used. 

Related vignette: https://cran.r-project.org/web/packages/dplyr/vignettes/in-packages.html#data-masking-and-tidy-selection-notes

(This can be done another day if the package doesn't do it already)

toph-allen: Thanks!

toph-allen:
> > * This is slightly different from similar functions like `get_group_members(client, group_guid)`. Reasons for this difference are:
> > 
> > Just passing in the `group_guid` is marginally simpler for the user for getting one group's content (just select the row, don't need to pull the column), and significantly simpler for getting multiple groups (just pass in a data frame with multiple rows). Landed upon this approach & reasoning rubber-ducking with @schloerke earlier last week.
> 
> Both `get_group_members()` and `get_group_content()` could move towards accepting both `src, groups` where `groups` is either a single `group_guid`, a `group_guid` vector, or a `groups` data frame. Given both methods return a data frame, it would extend nicely. (`get_group_members()` would need to add the `group_guid` / `group_name` for consistency w/ `get_group_content()`)

@schloerke How much do you think I should hesitate to change `get_group_members()` now because it breaks existing workflows? Like, should that change wait for a major release?

R/groups.R
toph-allen: Do you know why `lintr` doesn't seem to have a problem with some of the data-vars?

```r

  dplyr::transmute(parsed,
    group_guid = guid,
    group_name = name,
    .data$content_guid,
    .data$content_name,
    content_title,
    access_type,
    role = purrr::map_chr(
      .data$permissions,
      extract_role,
      principal_name = name
    )
  )
```

It seems fine with `content_title` and `access_type`. I'm adding `.data$` to them regardless for consistency & clarity, otherwise it looks like they are something different.

toph-allen: Apparently

```
Use of .data in tidyselect expressions was deprecated in tidyselect 1.2.0.
i Please use `"guid"` instead of `.data$guid`
```

toph-allen:
Marking this as ready for review. There are some areas that might be over-engineered! The function supports passing in bare `guid`s as well as the `groups` data frame. If passed in just GUIDs, it calls the `v1/groups` endpoints for each to look up the name.

Originally I thought we needed the name to filter the permissions list, but I mistakenly thought that the `principal_guid` was different from the `group_guid` (it is the same). Thus, we really only need the `name` to include in the output data frame. I still think this is much nicer than just including the GUID, though, so I'd like to keep it.

In the future (or possibly now?) I'll update `get_group_members` to work similarly in terms of arguments and data returned. I've held off on changing it now to avoid breaking changes, but if folks think it's fine breaking things to make them better, I'll update it now.

schloerke (approved):

jonkeane (commented):
At a higher level: do we need to have this dual-input setup where one can pass _either_ guids or a data.frame? What happens if the data.frame passed has data that is contradictory a guid and name in the same row but those don't match what's on the server?

R/groups.R
```diff
@@ -0,0 +1,205 @@
...
+    permissions,
+    ~ .x[["principal_guid"]] == principal_guid
+  )
+  if (length(matched) == 1) {
+    return(matched[[1]][["principal_role"]])
+  } else {
+    stop("Unexpected permissions structure.")
+  }
+  stop(glue::glue("Could not find permissions for \"{principal_guid}\""))
+}
```
toph-allen, line 205: New functions that comprise the implementations of the `get_group_content()`, factored out slightly for clarity of code and testability.

Notes for reviewers here, which should probably become code comments:

- The exported function, `get_group_content()`, handles input validation and normalization (to allow the use of both GUIDs and data frames), and calls `get_group_content_impl()` across all input groups.
- `get_group_content_impl()` gets the content for a single group from the guid and name. It isn't exported because it is the strictly less flexible core of `get_group_content()`.
- `extract_role()` is factored out for unit tests. Probably also useful elsewhere.

R/ptype.R
```diff
@@ -211,5 +211,41 @@ connectapi_ptypes <- list(
     principal_guid = NA_character_,
     principal_type = NA_character_,
     role = NA_character_
+  ),
+  group_content = tibble::tibble(
+    content_guid = NA_character_,
+    content_name = NA_character_,
+    content_title = NA_character_,
+    access_type = NA_character_,
+    permissions = NA_list_
```
toph-allen, line 220: For use with `parse_connectapi_typed()`

R/ptype.R
```diff
@@ -211,5 +211,41 @@ connectapi_ptypes <- list(
...
+      error = function(e) {
+        stop(glue::glue(
+          "Column `{col}` has type `{vctrs::vec_ptype_abbr(input[[col]])}`; ",
+          "needs `{vctrs::vec_ptype_abbr(required[[col]])}:`\n",
+          conditionMessage(e)
+        ))
+      }
+    )
+  }
+}
```
toph-allen, line 251: Added to allow us to easily validate data frames against a schema in functions, and to give decent user-interpretable error messages.

tests/testthat/_snaps/groups.md
```diff
@@ -0,0 +1,42 @@
...
+
+    Code
+      get_group_content(client, groups_df)
+    Condition
+      Warning:
+      Use of .data in tidyselect expressions was deprecated in tidyselect 1.2.0.
+      i Please use `"guid"` instead of `.data$guid`
+      Warning:
+      Use of .data in tidyselect expressions was deprecated in tidyselect 1.2.0.
+      i Please use `"name"` instead of `.data$name`
```
toph-allen, line 11: These warnings don't appear when I use the function from the command line. As far as I can tell, there are no other good options here:

- Just using `guid` with no qualifications gives lint errors
- Using `"guid"` fixes lint errors but changes the names of the resulting data frame to `"guid"` with quotation marks(!)

toph-allen (commented):
Added contextual comments for reviewers.

schloerke:
> > > * This is slightly different from similar functions like `get_group_members(client, group_guid)`. Reasons for this difference are:
> > > 
> > > Just passing in the `group_guid` is marginally simpler for the user for getting one group's content (just select the row, don't need to pull the column), and significantly simpler for getting multiple groups (just pass in a data frame with multiple rows). Landed upon this approach & reasoning rubber-ducking with @schloerke earlier last week.
> > 
> > 
> > Both `get_group_members()` and `get_group_content()` could move towards accepting both `src, groups` where `groups` is either a single `group_guid`, a `group_guid` vector, or a `groups` data frame. Given both methods return a data frame, it would extend nicely. (`get_group_members()` would need to add the `group_guid` / `group_name` for consistency w/ `get_group_content()`)
> 
> @schloerke How much do you think I should hesitate to change `get_group_members()` now because it breaks existing workflows? Like, should that change wait for a major release?

I missed this from earlier. My fault.

https://github.com/search?q=get_group_members+language%3AR+&type=code shows there isn't any usage outside of Posit. The shape of the function isn't really changing. The second argument would still be `guid` compatible. I would feel comfortable renaming the variable as the typical usage does not involve naming the variable.

But it seems fair to just try this new style moving forward and when the day comes, we can catch up `get_group_members()` then.

toph-allen:
> At a higher level: do we need to have this dual-input setup where one can pass _either_ guids or a data.frame? What happens if the data.frame passed has data that is contradictory a guid and name in the same row but those don't match what's on the server?

@jonkeane I think it's really nice to be able to pass in a data frame; it means that you can directly use the output of `get_groups()` or a `dplyr::filter()` call on that output. It feels nice and idiomatic to pass in a data frame, and using the data in that data frame to populate `group_name` and `group_guid` columns of the resulting data frame feels like a natural way to allow the function to operate on multiple groups at once, giving the user a lot of flexibility with the returned data. In the [Connect issue](https://github.com/rstudio/connect/issues/28849) about the cookbook I sketched out a few possible approaches, and the data frame feels the best. And allowing a vector of GUIDs to also be used as the input makes the function behave in a way that also is compatible with the existing idioms of the package (at a performance penalty, because in that case the `group_name` is looked up on the server.

As to your second question: If the user passes in an arbitrary data frame with `guid` and `name` columns that don't refer to groups on the Connect server (maybe accidentally a data frame of content items), then the API calls to look up the group content via the `guid` won't work and the function will return an error.

If they pass in a data frame that has real `guid`s from the Connect server but a nonsensical `name` column, those names would be used in the resulting data frame. I think that this is pretty unlikely, but it's definitely possible. They'd have to specifically construct that data frame using group `guid`s and some other text for a `name` column.

There are a few ways we could avoid this, but they all involve some level of compromise:

- Right now if the user passes in a vector of GUIDs, we look up the `group_name` for each GUID in a separate API call. This adds time and call overhead, but we could _only_ use the `guid` column of the data frame that's passed in and discard all other columns.
- We could remove `group_name` from the output, but this makes it harder to make sense of the results — even in the case of passing in one group GUID, but especially in the case of passing in multiple.

Another option, which I like less, would be to just implement this in a way that operates on a single group and does not include a `name` in the output.

toph-allen:
> > > > * This is slightly different from similar functions like `get_group_members(client, group_guid)`. Reasons for this difference are:
> > > > 
> > > > Just passing in the `group_guid` is marginally simpler for the user for getting one group's content (just select the row, don't need to pull the column), and significantly simpler for getting multiple groups (just pass in a data frame with multiple rows). Landed upon this approach & reasoning rubber-ducking with @schloerke earlier last week.
> > > 
> > > 
> > > Both `get_group_members()` and `get_group_content()` could move towards accepting both `src, groups` where `groups` is either a single `group_guid`, a `group_guid` vector, or a `groups` data frame. Given both methods return a data frame, it would extend nicely. (`get_group_members()` would need to add the `group_guid` / `group_name` for consistency w/ `get_group_content()`)
> > 
> > 
> > @schloerke How much do you think I should hesitate to change `get_group_members()` now because it breaks existing workflows? Like, should that change wait for a major release?
> 
> I missed this from earlier. My fault.
> 
> https://github.com/search?q=get_group_members+language%3AR+&type=code shows there isn't any usage outside of Posit. The shape of the function isn't really changing. The second argument would still be `guid` compatible. I would feel comfortable renaming the variable as the typical usage does not involve naming the variable.
> 
> But it seems fair to just try this new style moving forward and when the day comes, we can catch up `get_group_members()` then.

@schloerke I imagine that most of the usage of `connectapi` will probably be in private repos, but good point. I'll see how other discussion around this issue resolves and think about this more. My personality makes me want to follow semantic versioning fairly strictly, and I also don't want to increment a major version for a single parameter change in one function, so I kinda was thinking of coalescing changes like this into larger releases, but I take your point that few people will be calling `get_group_members(client, guid = "MY_GUID")`

jonkeane:
> @jonkeane I think it's really nice to be able to pass in a data frame; it means that you can directly use the output of `get_groups()` or a `dplyr::filter()` call on that output. It feels nice and idiomatic to pass in a data frame

If we only accepted GUIDs, you could still go from `get_groups()` to `dplyr::filter()` to `get_group_content()` you just need to `pull()` the GUID column.

>  and using the data in that data frame to populate `group_name` and `group_guid` columns of the resulting data frame feels like a natural way to allow the function to operate on multiple groups at once, giving the user a lot of flexibility with the returned data. In the https://github.com/rstudio/connect/issues/28849 about the cookbook I sketched out a few possible approaches, and the data frame feels the best. And allowing a vector of GUIDs to also be used as the input makes the function behave in a way that also is compatible with the existing idioms of the package (at a performance penalty, because in that case the `group_name` is looked up on the server.

I actually had missed in my first reading of the code that we are using the input data.frame to construct the output for group names. That's fishy — it opens the rest of the code up to having to deal with edge cases that we shouldn't (need to).

> As to your second question: ...

_nods_ yeah, those edge cases sound annoying to deal with. It might mostly work in most cases, but guarding against them is nontrivial (one example of this is in the PR itself: there's a non-trivial amount of guarding code that we have here, and if we push this forward we would likely need even more) 

> There are a few ways we could avoid this, but they all involve some level of compromise:

To be consistent: the approach of accepting data.frames also has compromises in the amount of guarding code we need to write + the edge cases we need to address when supporting that. Additionally, there is a documentation + user burden to having an argument that accepts two types and has slightly different behavior based on that. Sometimes it's worth it to do that anyway, but I don't think it is in this case.

Would it be possible to implement this PR as just accepting GUIDs? There are open questions about possible optimizations (getting group names) or alternative inputs (data.frames). But getting the simple case of GUIDs in a good state and merging that first would do us well here.

toph-allen:
Ok, I can do that.

R/groups.R
```diff
@@ -0,0 +1,190 @@
...
+
+  dplyr::transmute(parsed,
+    group_guid = guid,
+    group_name = permissions_df$principal_name,
+    .data$content_guid,
+    .data$content_name,
+    .data$content_title,
+    .data$access_type,
+    role = permissions_df$principal_role
+  )
```
jonkeane, line 189:
```suggestion
    tibble::tibble(
    group_guid = guid,
    group_name = permissions_df$principal_name,
    content_guid = parsed$content_guid,
    content_name = parsed$content_name,
    content_title = parsed$content_title,
    access_type = parsed$access_type,
    role = permissions_df$principal_role
  )
```

What about this? it avoids a call to dplyr entirely (and transmute is superseded anyway)

R/groups.R
```diff
@@ -0,0 +1,190 @@
...
+  }
+  parsed <- parse_connectapi_typed(res, connectapi_ptypes$group_content)
+
+  permissions_df <- purrr::map_dfr(
+    parsed$permissions,
+    ~ purrr::keep(
+      .x,
+      ~ .x[["principal_guid"]] == guid
+    )
+  )
```
jonkeane, line 179: This goes through the permissions (list of data.frames) column and returns a dataframe that is the matching rows from that list, yeah?

R/groups.R
```diff
@@ -0,0 +1,190 @@
...
+  validate_R6_class(src, "Connect")
+
+  purrr::map_dfr(
+    guids,
+    ~ get_group_content_impl(src = src, guid = .x)
+  )
+}
+
+#' @importFrom rlang .data
+get_group_content_impl <- function(src, guid) {
```
jonkeane, line 156: Do we still need this separate level of nested functions here still? I think it made more sense when there were a few more branches, but it feels unnecessary here. 

I also wonder if we can reduce the number of (practical) loops we're going through. We call this in a `purrr::map..` up above, and then there is at least one `purrr::map...` below which includes two anonymous functions which also practically go through each element checking for equality. These might all be necessary, but when I was trying to simplify this found the levels of nesting to be not-totally-obvious what was doing what where.

R/connect.R
```diff
@@ -687,6 +687,20 @@ Connect <- R6::R6Class(
...
+      self$GET(path)
+    },
+
+    #' @description Get the details for a group
+    #' @param guid The group GUID.
+    group_details = function(guid) {
+      path <- v1_url("groups", guid)
+      self$GET(path)
+    },
+
```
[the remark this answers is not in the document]
jonkeane: Is `group_details()` called anywhere?

jonkeane (changes requested):
Much more minimal, which is a great 💯 . A few comments about reducing some of the nesting of functions

R/groups.R
toph-allen: Hmm, I'm not sure. I found the distinction at this level to be helpful when thinking about the code — thinking about parsing each group's content into a data frame and then mapping that across `map_dfr` was helpful. But it would be possible to put this in an anonymous function call.

It isn't immediately obvious to me how to reduce the level of nestedness.

- This level is necessary, because we receive multiple GUIDs, and each one requires some level of processing.
- The inner processing is fairly nested because only the top level of the incoming JSON is parsed into a data frame, and the API returns a fairly nested structure for `permissions` (see below).

R/groups.R
toph-allen: `permissions` is a list of lists… of lists, hence the nestedness and the use of `purrr::keep`.

- At top level, each item is a content item i.e. a row of the data frame.
- The second level (i.e. within each each row), each item is a `principal` — potentially multiple per content item.
- In the third level, each item is a property of that principal permissions record (`principal_name`, `principal_role`, `principal_guid`, `principal_type`, etc.).

The API returns _all_ the principals for the content item, so we have to filter for the correct principal for each content item.

R/groups.R
toph-allen: Lol yes `transmute` is totally irrelevant here good catch 😂

R/groups.R
jonkeane: > This level is necessary, because we receive multiple GUIDs, and each one requires some level of processing.

Maybe we could make the name of this function make this clearer? If it were something like `get_one_group_content` that makes it a bit clearer that it's getting content for a single guid?

R/connect.R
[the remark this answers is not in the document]
toph-allen: It was previously, when I was looking up the group name for GUID-only invocations. But it's not now, so I've removed it.

R/groups.R
```diff
@@ -0,0 +1,188 @@
...
+#' @family groups functions
+#' @export
+get_group_members <- function(src, guid) {
+  validate_R6_class(src, "Connect")
+
+  res <- src$group_members(guid)
+
+  out <- parse_connectapi(res$results)
+
+  return(out)
```
jonkeane, line 105:
```suggestion
  parse_connectapi(res$results)
```

Two things: (1) we don't need to write this to `out` yeah? (2) We use implicit returns at the end of functions, right?

R/groups.R
```diff
@@ -0,0 +1,188 @@
...
+get_group_content <- function(src, guids) {
+  validate_R6_class(src, "Connect")
+
+  purrr::map_dfr(
+    guids,
+    ~ get_one_group_content(src = src, guid = .x)
+  )
+}
+
+get_one_group_content <- function(src, guid) {
```
jonkeane, line 154:
```suggestion
get_one_groups_content <- function(src, guid) {
```

Super minor: slightly more grammatical?

R/groups.R
```diff
@@ -0,0 +1,188 @@
...
+#' }
+#'
+#' @family groups functions
+#' @export
+get_group_content <- function(src, guids) {
+  validate_R6_class(src, "Connect")
+
+  purrr::map_dfr(
+    guids,
+    ~ get_one_group_content(src = src, guid = .x)
```
jonkeane, line 150:
```suggestion
    ~ get_one_groups_content(src = src, guid = .x)
```

NAMESPACE
```diff
@@ -147,6 +148,7 @@ importFrom(lifecycle,deprecated)
 importFrom(magrittr,"%>%")
 importFrom(rlang,"%||%")
 importFrom(rlang,":=")
+importFrom(rlang,.data)
```
jonkeane, line 151:
```suggestion
```

This looks like cruft from before — I suspect `devtools::document()` would do this too

jonkeane (approved):
Thanks for this! Looks good — a few very minor cleanups

NAMESPACE
toph-allen: Oh yes, I need to run it.

PR #339, opened 2024-11-27
fix: task returned from `content_render` missing `task_id`

### LINKED ISSUE #338: `poll_task` is trowing a 404 error because of changes in api reference
I used the documented example in [posit documentation](https://docs.posit.co/connect/cookbook/content/rendering-content/) and got a 404 error. 
Image

```r
library(connectapi)

CONTENT_GUID <- "34e425fb-179c-48e8-81bd-b334eb449f55"

client <- connect()
content <- content_item(client, CONTENT_GUID)
render_task <- content_render(content)

poll_task(render_task)
```
If you change this line
https://github.com/rstudio/connectapi/blame/f6093432582e1bed9ebcadbf613944cd691e5ead/R/deploy.R#L583

to
```r
  task_data <- con$task(task$task$id, wait = wait, first = first)
```

It solves the problem. I saw some changes to the tasks endpoint in august and I think this code was not updated as id parameter is now mandatory.

Thanks in advance for fixing this small bug !

toph-allen:
## Intent

Fix `poll_task` for tasks returned by `content_render`.

Fixes #338 

## Approach

The `POST variants/{variant_id}/render` endpoint returns an object that is shaped differently from the task returned by the `deploy` task, with the task ID named `id` and not `task_id`. Thus, when creating a `VariantTask` object, we need to rename the parameter.

I believe this was working in an initial draft of this functionality, but was changed in a subsequent PR that altered the return type for the function.

I added an integration test to verify `content_render` and the task it returns.

## Checklist

### DIFF
--- a/R/content.R
+++ b/R/content.R
@@ -1007,6 +1007,7 @@ content_render <- function(content, variant_key = NULL) {
     target_variant <- get_variant(content, variant_key)
   }
   render_task <- target_variant$render()
+  # render_task$task_id <- render_task$id
 
   VariantTask$new(
     connect = content$connect,
--- a/R/content.R
+++ b/R/content.R
@@ -1007,9 +1007,6 @@ content_render <- function(content, variant_key = NULL) {
     target_variant <- get_variant(content, variant_key)
   }
   render_task <- target_variant$render()
-  # Tasks returned from variant render endpoint look different from those
-  # returned by deploy.
-  render_task$task_id <- render_task$id
 
   VariantTask$new(
     connect = content$connect,
--- a/R/content.R
+++ b/R/content.R
@@ -1007,7 +1007,9 @@ content_render <- function(content, variant_key = NULL) {
     target_variant <- get_variant(content, variant_key)
   }
   render_task <- target_variant$render()
-  # render_task$task_id <- render_task$id
+  # Tasks returned from variant render endpoint look different from those
+  # returned by deploy.
+  render_task$task_id <- render_task$id
 
   VariantTask$new(
     connect = content$connect,
--- a/R/deploy.R
+++ b/R/deploy.R
@@ -580,7 +580,7 @@ poll_task <- function(task, wait = 1, callback = message) {
   code <- -1
   first <- 0
   while (!finished) {
-    task_data <- con$task(task$task$task_id, wait = wait, first = first)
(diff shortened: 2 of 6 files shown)


R/deploy.R
```diff
@@ -580,7 +580,7 @@ poll_task <- function(task, wait = 1, callback = message) {
   code <- -1
   first <- 0
   while (!finished) {
-    task_data <- con$task(task$get_task()$task_id, wait = wait, first = first)
+    task_data <- con$task(task$task$task_id, wait = wait, first = first)
```
nealrichardson, line 583: You don't technically need this change right? `get_task()` just yields `self$task`. (Not opposed to ripping out the unnecessary getters, but that seems better done in a separate refactor PR where you did them all.)

R/content.R
```diff
@@ -1007,6 +1007,9 @@ content_render <- function(content, variant_key = NULL) {
     target_variant <- get_variant(content, variant_key)
   }
   render_task <- target_variant$render()
+  # Tasks returned from variant render endpoint look different from those
+  # returned by deploy.
+  render_task$task_id <- render_task$id
```
nealrichardson, line 1012: I see this same logic in the Task `initialize` method: https://github.com/rstudio/connectapi/blob/main/R/deploy.R#L67-L69

Why doesn't that handle this for you already?

nealrichardson: Right, because of course `VariantTask` doesn't inherit from `Task` :/

I think the better fix would be to push this adapting into the `VariantTask` init method. Even better would be to look to fix the inheritance of all of these apparent (but not actual) Task subclasses, but that could be a followup--I'm not trying to derail a bugfix by tempting you into refactoring all the things.

R/deploy.R
toph-allen: Correct, I don't technically need it; happy to reverb if you think it's *bad* to change it here.

toph-allen: I agree, for things like this which are easily greppable, no need to do it piecemeal.

R/content.R
toph-allen: Yeah lol. But good point, I'll put it into `VariantTask` and rip out the other place that also does it.

I don't _think_ `ContentTask` needs it because I don't think there are any content APIs that return this form.

nealrichardson (approved):
Thanks for the quick and crisp fix!

PR #342, opened 2024-12-04
implement client credentials oauth flow

### LINKED ISSUE #344: implement client credentials oauth flow
for rstudio/connect/issues/28386, add a function to exchange a content-session-token for a client credential access token

christierney:
## Intent

Add oauth client credential token function to the package.

Fixes rstudio/connect#28386
Fixed #344 

## Approach

There was some discussion at https://github.com/rstudio/connectapi/pull/292#discussion_r1738978267 about whether it would be better to use a different function for client credentials vs viewer auth, or one function with different optional params. I opted for a separate function and followed the naming pattern from https://github.com/posit-dev/posit-sdk-py/pull/348/files

## Manual testing

From RStudio, installed development version of the library:

```r
remotes::install_github("rstudio/connectapi@chris-28386-cred-exchange")
```

and deployed this shiny app to local Connect:

```r
library(connectapi)
library(shiny)

ui <- fluidPage(
  textOutput("content_session_token"),
  textOutput("content_token"),
  )

server <- function(input, output, session) {


  content_session_token <- Sys.getenv("CONNECT_CONTENT_SESSION_TOKEN")
  output$content_session_token <- renderText(content_session_token)

  client <- connect(
    server = "http://localhost:3939",
    api_key = Sys.getenv("CONNECT_API_KEY"),
  )

  output$content_token <- renderText({
    resp <- get_oauth_content_credentials(client, content_session_token)
    resp$access_token
  })
}

# start the Shiny app
shinyApp(ui = ui, server = server)
```

Configured the content to use a valid service account and verified I can get a token:

Screenshot 2024-12-04 at 4 49 12 PM

## Checklist


R/get.R
```diff
@@ -629,6 +629,58 @@ get_oauth_credentials <- function(connect, user_session_token) {
...
+#'   # ... do something with `credentials$access_token` ...
+#'
+#'   "done"
+#' }
+#' }
+#'
+#' @return The OAuth credential exchange response.
+#'
+#' @details
+#' Please see https://docs.posit.co/connect/user/oauth-integrations/#obtaining-a-content-oauth-access-token
```
christierney, line 663: note: this doesn't exist yet but feels like a section we will add? Corresponds to the existing `#obtaining-a-viewer-oauth-access-token` section.

zackverham (approved):
LGTM but obviously want Toph to take a look still!

R/get.R
```diff
@@ -629,6 +629,58 @@ get_oauth_credentials <- function(connect, user_session_token) {
   )
 }
 
+#' Perform an OAuth credential exchange to obtain a content-specific OAuth
+#' access token.
+#'
+#' @param connect A Connect R6 object.
+#' @param content_session_token The content session token. This token can only
```
toph-allen, line 636:
```suggestion
#' @param content_session_token Optional. The content session token. This token can only
```

To go along with suggestion below. And I'd also adjust the prose a little too maybe :)

toph-allen (changes requested):
Looks good, but I think having it automatically retrieve the token from the environment in the default case is nicer — could you make that change?

R/get.R
```diff
@@ -629,6 +629,58 @@ get_oauth_credentials <- function(connect, user_session_token) {
...
+#' @return The OAuth credential exchange response.
+#'
+#' @details
+#' Please see https://docs.posit.co/connect/user/oauth-integrations/#obtaining-a-content-oauth-access-token
+#' for more information.
+#'
+#' @export
+get_oauth_content_credentials <- function(connect, content_session_token) {
+  validate_R6_class(connect, "Connect")
+  url <- v1_url("oauth", "integrations", "credentials")
```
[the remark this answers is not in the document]
christierney: thank you! this is super helpful.

R/get.R
zackverham: here's the in-flight draft of the user guide changes: 
https://github.com/rstudio/connect/pull/29116

The parallel actually ended up being `#obtaining-a-service-account-oauth-access-token`

toph-allen (approved):
Looks good! Just one small non-functional change, so approving. :) If you know you're gonna get that docs link in for this Connect release, I'm happy to merge now, or we can hold off on merging till that happens.

Thinking out loud here: Do you think it's worth renaming `get_oauth_credentials()` to `get_oauth_user_credentials()`? Could do that with a deprecation — keep the existing function around but do a deprecation warning and just have it call the renamed function. That way we have `get_oauth_content_credentials()` and `get_oauth_user_credentials()`. Happy to make that change after this merges in a separate PR.

toph-allen:
It looks like CI is failing because of something flaky, not for an actual reason, but if it continues to fail I'm happy to pull locally and poke around at it.

christierney:
re. naming: we'd also want to change the name in the python SDK which currently follows the same convention as this one. This also follows the convention in the Go oauth package. I do think it's a question worth raising, but probably separate from this PR.

R/utils.R
```diff
@@ -146,17 +146,15 @@ error_if_less_than <- function(using_version, tested_version) {
...
-      "You may be experience errors when using this functionality."
+      "You may experience errors when using this functionality."
     )
     warn_once(msg)
   } else if (comp < 0) {
     msg <- glue::glue(
-      "ERROR: This API requires Posit Connect version {tested_version} ",
-      "but you are using {using_version}. Please use a previous version ",
-      "of the `connectapi` package, upgrade Posit Connect, or review the API ",
-      "documentation corresponding to your version."
```
toph-allen, line 159: I changed the error messages.

- The phrase "this API requires…" doesn't really make sense. Users of the package will be thinking in terms of features or functions.
- The guidance in the error case was confusing. Using a previous version of `connectapi` would not help! If the feature doesn't exist on an old version of Connect, using an older `connectapi` will… just remove the `connectapi` references to that feature. Reviewing the API docs won't help either, because users won't see this error if they are directly calling the API, only if they're using existing functions.

christierney:
@toph-allen the tests don't need to change to accommodate the version check? What version of Connect do tests see?

christierney:
oh wait, toph already explained this to me in slack:

> Ah, the tests don’t fail because the mock server version is NA (i.e. not exposed), which just results in a warning, not an outright error. I thought that might be the case, probably I adjusted it for this very reason at some point

toph-allen (approved):

PR #343, opened 2024-12-04
feat: use v1 api in `get_jobs()`

### LINKED ISSUE #340: Migrate `get_jobs()` to v1 API
`connectapi` currently calls the unversioned API when using `get_jobs()`. Migrate to the v1 API, calling the unversioned API as a fallback. The returned data types differ between the two APIs, so come up with a way to return a consistent data type.

toph-allen:
## Intent

Switch to using the `v1` API for `get_jobs()`.

Fixes #340 

## Approach

The data returned by the old unversioned endpoint and the v1 endpoint (which has been around since 2022) differs. I adjusted the `ptype` to match the new endpoint, which mostly processes the old data correctly. To support this, I added the following:

- Now there is an option to use the `ptype` column order in `parse_connectapi_typed()`, to make the returned data nice and consistent looking. :)
- Manually compute `status` (new) from `finalized` (old). We cannot fully compute `status`. The old `finalized == FALSE` corresponds to `status ==  0`, but `finalized == TRUE` corresponds to `status`es `1` and `2`, so in that case `status` is `NA`.
- Added a function to let `vctrs` cast from integer to character.

Tests added:

- A test using `httptest`, which uses real responses captured from the old and new endpoints for the same piece of content. Checks that common columns are identical, and manually checks the `status` column for each.

## Checklist

- [ ] Does this change update `NEWS.md` (referencing the connected issue if necessary)? **NEED**


R/parse.R
```diff
@@ -27,14 +27,19 @@ make_timestamp <- function(input) {
...
   # Given a prototype, ensure that all columns are present and cast to the correct type.
   # If a column is missing in .data, it will be created with all missing values of the correct type.
   # If a column is present in both, it will be cast to the correct type.
   # If a column is present in .data but not in ptype, it will be left as is.
   for (i in names(ptype)) {
     .data <- ensure_column(.data, ptype[[i]], i)
   }
+
+  if (order_columns) {
+    .data <- .data[, names(ptype), drop = FALSE]
```
schloerke, line 40: We should try to keep the names of data as well, in case it has more columns than ptype.

Untested!

```suggestion
    .data <- .data[, unique(c(names(ptype), names(.data))), drop = FALSE]
```

R/content.R
```diff
@@ -93,11 +93,25 @@ Content <- R6::R6Class(
     get_dashboard_url = function(pane = "") {
       dashboard_url_chr(self$connect$server, self$content$guid, pane = pane)
     },
-    #' @description Return the jobs for this content.
+    #' @description Return the jobs for this content
```
schloerke, line 96: Restoring `.` to match other `@description` fields that contain a single sentence.

```suggestion
    #' @description Return the jobs for this content.
```

schloerke (approved):
NEWS entry worthy?

Two small things, then good to merge!

R/parse.R
toph-allen: In a newer pull request I changed this argument name to `use_ptype_cols` to make this part of the behavior more clear (that behavior is used to drop the `guid` column). But perhaps it's better to decompose those two behaviors (reordering, keeping only ptype columns).

Issue #305, opened 2024-09-12
RFC: A new user experience for connectapi


nealrichardson:
Follow pattern of posit-sdk: base Resources and Resource classes, containers and object inheriting from them.

`Resources` (collections) have an `as.data.frame()` method to return a data frame (like the `get_content()` etc. functions do now). Otherwise they are just interfaces.

Assuming R6, you could chain things together with `$`, like `client$users$get_by_id(1)$update(first_name = "Jane")`, but that's not idiomatic R. So we can add some helper functions like this:

```r
users <- function(object) {
  object$users
}

# actually, update is a S3 generic in stats, so you'd do this with methods
update <- function(object, ...) {
  object$update(...)
}
```

Then you can pipe.

```r
client |>
  users() |>
  create(first_name = "John", last_name = "Doe")

client |>
  users() |>
  get_by_id(1) |>
  update(first_name = "Jane")

client |>
  content() |>
  as.data.frame()

client |>
  content() |>
  get_all() # returns a list
```

Defer all requests until actually needed. I.e. `client |> users()` doesn't request all users because you don't need them if you're then going to do `create()` or `get_by_id()`.

The reason to use R6 would be for memoizing the requests. For example, if you do `client |> users() |> get_by_id(1)`, we still haven't made any HTTP requests. Only if you do `client |> users() |> get_by_id(1) |> first_name()` would we actually make the request, then cache the response data in the object, so that if we then did `last_name()`, we wouldn't have to request again. And if we instead did `client |> users() |> get_by_id(1) |> update(first_name == "Jane")`, we would only make the `PATCH` request, no `GET`. 

We could also handle this caching at the HTTP layer, and then we wouldn't necessarily benefit from R6. https://enpiar.com/r/httpcache/ is a package I wrote for this a while back that we could use.

toph-allen:
@nealrichardson This issue proposes an update to the way we handle the Connect client: https://github.com/posit-dev/connectapi/issues/359

toph-allen:
I was thinking about this while thinking about naming for functions in #356. I'm going to use that workflow as a way to think about some of the features of that implementation.

[...]

PR #356, opened 2024-12-30
feat: add `get_log(job)`

### LINKED ISSUE #341: Migrate `get_job()` (singular) to v1 API
Similarly to #340, the singular `get_job()` uses an unversioned API and should move to the v1 endpoint.

The unversioned endpoint currently in use returns the `stdout` and `stderr` of the process. This is the only reason I imagine a user would want to use this function over the `get_jobs()` function.

The v1 endpoints for getting a single job just returns the same data as `get_jobs()`. There are two additional endpoints, though, that return the logs and any error message for a given process.

So `get_job_logs()` and `get_job_error()`

Perhaps return the logs as a data frame, with each message on a row, and separate columns for timestamp, data, and source (`stdout`, `stderr`).

Still run into the problem of having to pass in both a content item and a job key with the current approach, e.g.:

```r
CONTENT_GUID <- "guid"

client <- connect()
item <- content_item(CONTENT_GUID)
jobs <- get_jobs(item)
key <- jobs$key[1]
logs <- get_job_logs(item, key)
```

If `get_jobs()` returned a list of job class objects, we could just pass in the single job (the job object could contain a reference to its parent content item). But in the current mode of returning job data as a data frame, you get to, well, interact with that data with data frame semantics, which is really nice and feels like second nature to me (e.g. `dplyr::filter` etc.).

toph-allen:
## Intent

The initial intent of this PR was to move `get_job()` to use the v1 API.

However, that endpoint returns the exact same data as `get_jobs()`, so it seems kinda redundant. Is there any point at which you'll have a job key where you _haven't_ called `get_jobs()`? Not right now, I don't think.

The old unversioned `get_job()` function also returned the log for the job, and this seems to be the main use case for the singular function, so I instead added a new function, `get_log()`, which gets the job log.

Fixes #341

## Approach

~~The new function is `get_log(content_item, key)`. It's a little clunky to use, because the API requires the content GUID as well as the job key.~~

~~An alternate approach would be to add a `Job` R6 class. This would also probably also require a different method / signature for things like `terminate_jobs()`, because you'd wanna keep the existing implementation but also allow you to call `terminate_jobs(job_class)` (although that plural function name suggests that it takes a list of jobs, so maybe it would just call for a `terminate_job(job_class_object)` function.~~

- Adds a `get_job_list(content)` function which returns a list of jobs for a content item. Each job is just a list containing the API output, augmented with the `app_guid` (which is required for subsequent job requests but not contained in the data returned) and the Connect `client`.
- Adds a `get_log(job)` function which returns the job log for the job object as represented by that list item.

```r
client <- connect()
content <- content_item(CONTENT_GUID)
jobs <- get_job_list(content)
log <- get_log(jobs[[1]])
```

## Checklist


R/parse.R
```diff
@@ -169,8 +169,8 @@ parse_connect_rfc3339 <- function(x) {
     # Times with and without offsets require different formats.
     format_string <- ifelse(
       grepl("Z$", .x),
-      "%Y-%m-%dT%H:%M:%SZ",
-      "%Y-%m-%dT%H:%M:%S%z"
+      "%Y-%m-%dT%H:%M:%OSZ",
+      "%Y-%m-%dT%H:%M:%OS%z"
```
jonkeane, line 173: Was this a bug we hadn't caught here? I see the addition of a test (which is great!) but just curious if this has been wrong and you just caught it here or if something deeper is going on?

R/content.R
```diff
@@ -747,6 +747,31 @@ terminate_jobs <- function(content, keys = NULL) {
...
+#'
+#' @return A data frame with the requested log. Each row represents an entry.
+#'
+#' - `source`: `stdout` or `stderr`
+#' - `timestamp`: The time of the entry.
+#' - `data`: The logged text.
+#'
+#' @family job functions
+#' @family content functions
+#' @export
```
jonkeane, line 765: Adding literal `@examples` here is complicated (though could be done with httptest!) would you mind writing one in the comment here (or on the main PR), what is the code someone would write to use this?

jonkeane (commented):
Thanks, a few questions inline.

> An alternate approach would be to add a Job R6 class. This would also probably also require a different method / signature for things like terminate_jobs(), because you'd wanna keep the existing implementation but also allow you to call terminate_jobs(job_class) (although that plural function name suggests that it takes a list of jobs, so maybe it would just call for a terminate_job(job_class_object) function.

Could you say more about this? What would the benefits of this approach be? What about the downsides? It might also be helpful here to write the code that someone using this would write if this were the case and compare it to the code that one would right with this PR as it stands.

R/parse.R
toph-allen: Neither — `%S` was valid for most of the timestamps Connect sends, but logs contain decimal second precision, so to properly parse the log data from the `v1` endpoint, we need `%OS`. The existing tests confirm that this change preserves compatibility with timestamps lacking decimal seconds.

R/content.R
toph-allen: Yep, this was on my list of things to do, thanks for the reminder.

toph-allen: Noticing that the other jobs functions also lack examples (which is why I didn't remember to add one when I was writing this documentation). Going to add those as well.

toph-allen:
@jonkeane Can I get a re-review when you have a sec?

R/content.R
```diff
@@ -747,6 +761,49 @@ terminate_jobs <- function(content, keys = NULL) {
...
+#' - `timestamp`: The time of the entry.
+#' - `data`: The logged text.
+#'
+#' @examples
+#' \dontrun{
+#' client <- connect()
+#' item <- content_item(client, "951bf3ad-82d0-4bca-bba8-9b27e35c49fa")
+#' jobs <- get_jobs(item)
+#' key <- jobs$key[1]
+#' log <- get_job_log(item, key)
```
jonkeane, line 786: What would it take to live in a world where we could do something like:

```
client <- connect()
item <- content_item(client, "951bf3ad-82d0-4bca-bba8-9b27e35c49fa")
jobs <- get_jobs(item)
log <- get_log(jobs[1]) or log <- get_log(jobs[[1]]) 
```

It feels a little clunky that you need to keep track of both ID and pass that in.

This might also be similar to what I was thinking might be proposed when I asked https://github.com/rstudio/connectapi/pull/356#pullrequestreview-2526510437 sorry if I missed the answer to this elsewhere.

toph-allen: I agree, it's clunky! The solution would be to implement a `Job` class which keeps track of its parent content item. 

You're right, this is what you asked in that review. I responded to code-level questions you had but not the PR-level question — here are some thoughts about tradeoffs.

The major benefit is, of course, that a `Job` class would make this much easier to use, as the `Job` class could keep track of its reference.

Tradeoffs and some thoughts on them include:

## Compatibility

`get_jobs()` has been around for four and a half years ([ref](https://github.com/rstudio/connectapi/commit/7ceb01f0a24ee520295bd73a1404b74e138c2ab9)). I updated it in the last version to use the `v1` API, and the data it returns changed, but the shape is similar. Implementing a `Job` class would be a much more major change to code that's been around for a while.

Terminate `terminate_jobs()`, which was added recently takes a `Content` object and optional `keys`, and we might want to update it to function logically with a set of `Job` objects without changing its signature too badly. Thinking about it more, I think I lean towards keeping `terminate_jobs()` as a function primarily for terminating all the jobs of a `Content` object and adding a `terminate_job()` function to handle individual job objects.

## Working with R6 objects: selection / filtering

This gets kinda at the heart of some of the friction I feel with R6 objects — I've been thinking about this today as I'm working on the `Environment` implementation. In our examples, where we write `jobs[1]` or `jobs[[1]]`, our customers actually probably want to perform some kind of filtering so they're only operating on a subset of returned jobs, or finding a job that matches some criteria that are more complex than "first in the list".

Currently, `get_jobs()` returns a data frame, and in similar examples in the Cookbook we use `dplyr` for filtering. Switching to a `Job` class would switch us from data frame-like to list-like paradigms. This is a little more clunky — you have to use `Filter()` or `purrr::keep()`.

If we implemented the `Job` class following existing patterns in `connectapi`, the job data would be located on a `$job` field. With that implementation, you'd use something like the following to filter for jobs with `status == 1`:

```r
jobs <- get_jobs(content_item)

# Data frame, dplyr
target_jobs <- dplyr::filter(jobs, status == 1)

# Data frame, base R
target_jobs <- jobs[jobs$status == 1, ]

# List following existing `connectapi` patterns, purrr
target_jobs <- purrr::keep(jobs, ~ .x$job$status == 1)

# List following existing `connectapi` patterns, base R
target_jobs <- Filter(function(x) x$job$status == 1, jobs)
```

There are two problems: (1) making users access relevant fields under a `$job` property, and (2) filtering objects in lists feels less idiomatic.

I've been solving problem (1) by structuring the class to allow users to directly query the data at the top level. (I've done this by storing the underlying list of fields in a `$data` property. I've tried using active binding for each field, and also redefining `[[` and `$` to check `$data` before calling `get(name, envir = self)`. Both work.)

This gives us the following, which is better:

```r
jobs <- get_jobs(content_item)

# purrr
target_jobs <- purrr::filter(jobs, ~ .x$status == 1)

# Base R
target_jobs <- Filter(function(x) x$id == 1, envs)
```

This doesn't solve problem (2), which is just that for filtering from a set of jobs, my first impulse is definitely to want to filter them like a data frame.

What do you think? Does `purrr::keep()` feel simple enough for the increased simplicity of function calling?

One hare-brained idea I had was to have a `Jobs` class to represent a data frame of jobs, this allowing you to filter with data frame / vector filter/selection paradigms. But that doesn't really help with simplicity of calling a single `get_job_log()` function — like, each function would need to handle the case of multiple jobs being passed in.

Perhaps I'm overthinking this, and using `purrr::keep()` feels natural and simple.

jonkeane: > I guess we could have a ResourceCollection wrapper class that provided a more data-frame-like interface to collections of objects.

We could even do this without the classes too, yeah? Or with very bare classes for now as we build up + explore what those look like and if that's the right pattern too. If what was returned[1] was a list of objects, and each of those objects could be passed into `get_log()` and it did the right thing you wouldn't even need to have any classes/methods (of course, it makes sense to do that, and maybe that's the right thing in the end, but from a practical perspective of "here's the code I would love someone to be able to use what's the thing that gets me there and then let's me explore what underlying structure makes building / maintaining that easiest" doesn't *require* that we start with a pre-planned class structure.)

[1] I hear you when you say that `get_jobs()` has been around a long time and there's some change management we would need to do to update that. If that's the case, maybe we make a `get_jobs_2()` (I'm intentionally giving it a bad name, don't pick that!) that returns what we want. We can use `get_jobs()` internally (or not) but we can create something new if that's what we need to get to a place we want.

toph-allen: Hmm, perhaps I misunderstand your meaning here, but if the thing that is returned is a list of objects, then… I feel like those objects are at least a little more usable/tractable if they are classes. But perhaps not!

My thought process goes: Right now, `content$jobs()` returns a bare list of lists, which is processed into the data frame by `get_jobs()`. (This is an established pattern in `connectapi` and I've just been following it.) But this data would need to be augmented with at least the content GUID, so there'd need to be another function to wrap it. And if we're returning objects that are not just the bare lists returned by the API, then at least creating a minimal class makes it more understandable what that data is, makes it a little more documented or clear, both to users and to our future selves.

Relatedly, exploration of a minimum-viable-class is in this draft PR for execution environments: https://github.com/rstudio/connectapi/pull/357/files

toph-allen: @jon Something we didn’t think about yesterday using the list for the get_job_log() function is — as currently architected, if we remove the content argument, we need to add a client argument, because content is currently how the client is passed in.

This could be remedied by:

(1) Having a `Job` class that contains a reference to the client.
(2) Introducing the paradigm of a default client object that I've discussed here and there in the past. I wrote this proposal up into an issue [here](https://github.com/posit-dev/connectapi/issues/359) — our discussion yesterday makes me feel more strongly that this approach is beneficial, because it would free us up to implement lighter-weight solutions like your suggestion without needing to explicitly pass the client into every function that doesn't use an R6 class.

Neither of those changes feel like they are within scope for this PR, though.

R/content.R
```diff
@@ -747,6 +769,52 @@ terminate_jobs <- function(content, keys = NULL) {
...
+#' - `timestamp`: The time of the entry.
+#' - `data`: The logged text.
+#'
+#' @examples
+#' \dontrun{
+#' client <- connect()
+#' item <- content_item(client, "951bf3ad-82d0-4bca-bba8-9b27e35c49fa")
+#' jobs <- get_job_list(item)
+#' target_job <- jobs[[1]]
+#' log <- get_job_log(target_job)
```
jonkeane, line 800: This definitely does read nicer to me. Maybe simplified with:

```
#' log <- get_job_log(jobs[[1]])
```

toph-allen: I intentionally included that as a separate line to call it out a bit more clearly (by itself, with that variable name, that line says "this item is a job" rather than just "this item is an argument you pass to this function").

toph-allen: But I made the change. Can I get a re-review?

tests/testthat/test-content.R
```diff
@@ -270,6 +270,44 @@ test_that("get_jobs() using the old and new endpoints returns sensible results",
...
+        remote_id = NULL, app_id = "52389", variant_id = "0",
+        bundle_id = "127015", start_time = "2024-11-15T17:03:00Z",
+        end_time = "2024-11-15T17:06:23Z", last_heartbeat_time = "2024-11-15T17:06:20Z",
+        queued_time = NULL, queue_name = NULL, tag = "run_app",
+        exit_code = 0L, status = 2L, hostname = "dogfood02",
+        cluster = NULL, image = NULL, run_as = "rstudio-connect",
+        app_guid = "8f37d6e0", client = client
+      )
+    )
+  })
```
jonkeane, line 305: I wonder if it would be good to test a handful of these properties instead of all of them at once. Given the code that is in `get_job_list`, we probably don't have to worry that there's shuffling of the values there (where given the mocks you get a mix of stuff from 40793542 and 39368207 there. So maybe it's "just" testing the `id`, `client`, and `guid` bits (or which ever are the important bits for jobs+logs)? That would be kind(er) to future-you when these change rather than a big diff of a bunch of things that all changed at once, we have a smaller one.

tests/testthat/test-content.R
```diff
@@ -324,3 +362,34 @@ test_that("an error is raised when terminate_jobs() calls a bad URL", {
...
   })
 })
+
+test_that("get_job_log() gets job logs", {
+  with_mock_api({
+    client <- Connect$new(server = "http://connect.example", api_key = "not-a-key")
+    item <- content_item(client, "8f37d6e0")
+    job_list <- get_job_list(item)
+    job <- purrr::keep(job_list, ~ .x$key == "mxPGVOMVk6f8dso2")[[1]]
+    log <- get_job_log(job)
```
jonkeane, line 372: `key == "mxPGVOMVk6f8dso2"` is finding a specific job from the fixtures that you then assert about later, yeah? A comment might be nice that you're picking that one so that you can assert the shape of the data down below.

R/content.R
```diff
@@ -621,15 +621,21 @@ content_ensure <- function(
 #'
 #' @param content A Content object, as returned by `content_item()`
 #'
-#' @return A data frame with a row for each job, with the following columns:
+#' @return
+#'
+#' - `get_jobs()`: A data frame with a row representing each job.
+#' - `get_job_list()`: A list with each element representing a job.
```
jonkeane, line 627: If we include this, we should do it elsewhere, but should we talk a little bit about this difference in the documentation? Where to use one or the other?

R/content.R
```diff
@@ -747,6 +769,51 @@ terminate_jobs <- function(content, keys = NULL) {
...
+#' - `timestamp`: The time of the entry.
+#' - `data`: The logged text.
+#'
+#' @examples
+#' \dontrun{
+#' client <- connect()
+#' item <- content_item(client, "951bf3ad-82d0-4bca-bba8-9b27e35c49fa")
+#' jobs <- get_job_list(item)
+#' log <- get_job_log(jobs[[1]])
+#' }
```
jonkeane, line 800: Is one able to use `get_job_log()` with the output of `get_jobs()`? I think the answer is no — that's ok, but we should probably flag that.

R/content.R
```diff
@@ -747,6 +769,51 @@ terminate_jobs <- function(content, keys = NULL) {
...
+#' Get Job Log
+#'
+#' Get the log output for a job. Requires Connect 2022.10.0 or newer.
+#'
+#' @param job A job, represented by an element from the list returned by `get_job_list()`.
+#' @param max_log_lines Optional. An integer indicating the maximum number of
+#' log lines to return. If `NULL` (default), Connect returns a maximum of 5000
+#' lines.
+#'
+#' @return A data frame with the requested log. Each row represents an entry.
```
jonkeane, line 788: This is slightly dependent on the answer to the question in another comment (sorry I can't seem to link it while the review is pending):

> Is one able to use get_job_log() with the output of get_jobs()? I think the answer is no — that's ok, but we should probably flag that.

If this function only works with our new approach, I wonder if this should actually return a list rather than a data.frame? Then we have two paths: the old(er) data.frame-centric approach and the new(er) list-based approach. Keeping the paths separate there might help us explore and see how those interfaces work with each other.

jonkeane (commented):
Thanks for this. The list-based approach looks interesting and a promising direction to try out. 

One question I have is how much should we flag this tension / split to folks, especially (also commented below in the right place in the code):

> Is one able to use `get_job_log()` with the output of `get_jobs()`? I think the answer is no — that's ok, but we should probably flag that.

Having two separate paths a list-based one and a data.frame-based one is totally fine, but if we do that we should embrace it + document it.

R/content.R
toph-allen: I don't think the data returned by the log endpoint makes sense to return as a list — it doesn't feel necessary to draw this dichotomy to me. Maybe this is because the log returned feels more like _data_ rather than a representation of a _server object_, and the data frame is a much more useful representation of its data than a list.

The data frame with rows representing log entries is really useful — a user can filter log entries by `source` or `timestamp`, and extract `log$data` to get the text of the lines for the resulting entries. Returning it raw from the API, with each list item representing a log line… really, the best way to do something useful for it would be to transform it into a data frame.

If we really cared about making _everything_ a list-like object, we could return a list containing `job_key`, `app_guid`, and `log`, with the latter field containing the data frame of log data.

R/content.R
toph-allen: It would be reasonable to talk about this elsewhere. This might be a good place to talk about it, or it might make sense in the "details" section of this documentation.

While we're working on fleshing out the structure around this dichotomy.

I think we also might want to add a `terminate_job()` function that takes a single job object, analogous to `get_job_log()`. The current `terminate_jobs()` function takes a content item and an optional job key.

---

Aside — the reason I haven't used `get_log()` for this function is because Connect has multiple types of logs (e.g. audit, server) and

R/content.R
```diff
@@ -747,6 +769,51 @@ terminate_jobs <- function(content, keys = NULL) {
   res_df
 }
 
+#' @rdname get_jobs
+get_job_list <- function(content) {
```
toph-allen, line 773: @jonkeane Thinking it would be better to just call this `jobs()`, which would align it with https://github.com/posit-dev/connectapi/issues/305.

In other words — if a function returning a list-like collection of a Connect server objects is an evolutionary transition fossil on the way to an API like that described in #305, we should name it like that.

toph-allen: I actually misremembered the proposal when I was writing that comment on Sunday. The actual resource class is not the thing which could be accessed in list-like ways. That proposal uses a function `get_all()` (called on the resource class) to get a list of objects.

I think I'd actually want to avoid using a `jobs()` function name to leave it unoccupied for a resource class, should one be implemented. I was trying to think about whether it's reasonable to have a resource class that could be addressed like a list, but that's out of scope.

R/content.R
jonkeane: If it's more natural to leave if as a data.frame, that's fine

R/content.R
toph-allen: already mention that in the description for the `job` param to `get_job_log()`. You commented here — does that indicate you think there should be, like, a comment in the code block here?

R/content.R
toph-allen: In practice, having a function called `jobs()` that returns a list and a function called `get_jobs()` that returns a database is less intuitive than frustratingly wordy function names.

tests/testthat/test-content.R
toph-allen: I did as you suggested here.

tests/testthat/test-content.R
toph-allen: I don't fully understand what you mean here, but I added a clarifying comment.

R/content.R
toph-allen: I wrote a paragraph in the `Details` section of this functions documentation to get at the distinction. When you say "we should do it elsewhere" did you mean "outside of the function documentation" or "outside of the `@return` section"?

I want to add a `terminate_job()` function in a small follow-on PR that also takes job objects. That's another example to add to this documentation that helps to get at the distinction. The current `terminate_jobs()` function is built more for the data frame paradigm.

toph-allen:
@jonkeane I made some minor documentation changes responding to your review, but `git` pull/push operations are not working (on any repo) and I'm going down a rabbit hole trying to fix. :|

R/content.R
jonkeane: Ah no, I see it now. I missed it on first read. Maybe that means it should be highlighted more (especially given the intentional split we are making here), or maybe it means I need better glasses. I'll leave it up to you if you want to call that out more highly (like in the body text of help). 

FWIW, I commented here because it's where I thought about it looking at the examples and missed it in the argument which would have been a better place to comment it on. I might be reading slightly too much into word choice, but I agree with the (subtle) skepticism I'm reading about it being in a comment in the example as a way to highlight that.

toph-allen: I made a minor change to the documentation to mention this in more places and make it slightly more prominent. I believe these were the only issues you had in your review — as such I requested re-review.

But I think this gets at the major sticking point for me with the "it's just a list" approach — which is that there _is_ no good way to go from the data frame returned by `get_jobs()` to the object needed by `get_job_log()`. Without this, I think it's likely that we haven't *fully* thought through our users' workflows, and are building something that looks good in minimal examples but doesn't provide good ergonomics beyond that.

I'm fine merging this PR as is — documenting that limitation — I think it's best to move forward here. But this is something that I've been thinking about e.g. in #305.

---

One thought I had as a stepping stone in that direction, which almost works with what `connectapi` currently has, is a `job()` function that takes a job key and returns the server data (the stand-in for a "job" class). 

The workflow I have in mind is:

```r
client <- connect()

client |>
  content_item(GUID) |>
  get_jobs() |> # data frame
  slice_max(end_date) |> # get the most recent
  pull(key) |>
  job() |> # construct a job object — in this case just a list
  get_job_log()
```

But `job()` would need the client, and that's so far _always_ the first argument, so this workflow would need to break that paradigm. This is why I've been thinking about #359 (default client as a trailing arg). This would require a surplus HTTP request but that seems fine.

Another option that might actually work is just piping a single-row data frame to `as.list()` — except that the data frame lacks a list-column of `client` objects, i.e. it's not an exact match for the data from `get_job_list()`.

```r
client <- connect()

client |>
  content_item(GUID) |>
  get_jobs() |> # data frame
  slice_max(end_date) |> # get the most recent
  as.list() |> # wouldn't actually work right now
  get_job_log()
```

toph-allen:
Merging without approval after checking in with @jonkeane.

PR #361, opened 2025-01-10
fix: `swap_vanity_urls()` works predictably with unset vanity URLs

### LINKED ISSUE #360: `swap_vanity_url(content1, content2)` assigns junk URLs if one content item has no vanity set
This bug was discovered while building a "Swap Vanity URLs" Connect app.

If one of the content items has no vanity URLs, instead of receiving no vanity URL, the other content item will receive a junky placeholder vanity URL. It looks like this function may have been implemented before it was possible to delete vanity URLs.

Image

toph-allen:
## Intent

Update `swap_vanity_url()` with numerous fixes and changes. Changes include:

- When `swap_vanity_url()` was called on a piece of content without a vanity URL, the content item set to receive that would instead receive a randomized placeholder URL. Now the function works as intended when one content item has no vanity URL: the other content item receives an unset vanity URL!
- Better error handling. Previously, errors were not handled gracefully, so if the caller did not have permission to modify the second content item, the first content item was left with its randomized placeholder URL. Now, errors are called out correctly and any changes made are rolled back.
- Names! It is now called `swap_vanity_urls()` because you're swapping two vanity URLs between two content items. The arguments are now called `content_a` and `content_b` because the old argument names, `from` and `to`, don't make sense when vanity URLs move from and to both items. The old function name remains but is deprecated, and just shims argument and return object names to the new function.

Fixes #360

## Approach

Rewrote the function's internal workflow. Pseudocode:

1. Get both content item's current vanity URLs.
2. Delete each content item's vanity URLs, each inside a tryCatch statement. If `content_a` fails, stop and err. If `content_b` fails, restore `content_a`'s original vanity URL and err.
3. Assign each content item the other's URL. We now know these operations will succeed, because `delete` and `set` have the same permissions.

Perhaps a better approach would be to call `set_vanity_url(content_a, vanity_b, force = TRUE)` in 2., as this doesn't risk the (admittedly very unlikely) case that another piece of content snipes the URL before it's reassigned. But this actually introduces more complexity:
- We'd have to do conditional logic to know if we needed to `set` or `delete` in this step.
- Because `force = TRUE` also modifies `content_b`'s vanity URL, if we lacked permission to modify `content_b`, that operation would fail — but _only_ if `content_b` initially had a vanity URL (if its vanity was unset, we'd just be unsetting `vanity_a`'s URL). So the workflow for rollbacks would be much more complex, and harder to reason about / test / maintain.

## Checklist


R/deploy.R
```diff
@@ -464,14 +464,14 @@ delete_vanity_url <- function(content) {
   error_if_less_than(con$version, "1.8.6")
   guid <- content$get_content()$guid
 
-  con$DELETE(v1_url("content", guid, "vanity"))
+  con$DELETE(v1_url("content", guid, "vanity"), parser = "parsed")
```
toph-allen, line 467: So that it returns an error if it receives an HTTP error status.

R/deploy.R
```diff
@@ -501,55 +501,80 @@ get_vanity_url <- function(content) {
...
-    if (!is.null(to_vanity)) {
-      set_vanity_url(from_content, to_vanity)
+    if (!is.null(vanity_b)) {
+      set_vanity_url(content_a, vanity_b)
     }
-
-    from_vanity <- get_vanity_url(from_content)
-    to_vanity <- get_vanity_url(to_content)
+    vanity_a <- get_vanity_url(content_a)
+    vanity_b <- get_vanity_url(content_b)
```
tdstein, line 548: Nice, this is a good idea. The server does mutate the submitted vanity url in some cases (i.e., leading '/').

tdstein (approved):
Logical ✅ 
Test coverage ✅
Documentation ✅  

Looks good to me!

PR #363, opened 2025-01-17
feat: add support for visitor api key

### LINKED ISSUE #362: Add support for viewer API key
Add support for getting a viewer API key when running on Connect.

The Python implementation is here: https://github.com/posit-dev/posit-sdk-py/pull/372

The core handshake is most easily seen in this excerpt from a Python test fixture:

```python
def register_mocks():
     responses.post(
         "https://connect.example/__api__/v1/oauth/integrations/credentials",
         match=[
             responses.matchers.urlencoded_params_matcher(
                 {
                     "grant_type": "urn:ietf:params:oauth:grant-type:token-exchange",
                     "subject_token_type": "urn:posit:connect:user-session-token",
                     "subject_token": "cit",
                     "requested_token_type": "urn:posit:connect:api-key",
                 },
             ),
         ],
         json={
             "access_token": "viewer-api-key",
             "issued_token_type": "urn:posit:connect:api-key",
             "token_type": "Key",
         },
     )
```

In `connectapi`, the client is created with a `connect()` function, which gets the server URL and API key from environment variables by default. Perhaps the best workflow is adding an optional `user_session_token` argument to that function that, when provided, returns a viewer client object. Something like:

```r
library(shiny)
library(connectapi)

ui <- fluidPage(
  "Publisher API key",
  verbatimTextOutput("default_api_key"),
  "Viewer API key",
  verbatimTextOutput("viewer_api_key")
)

server <- function(input, output, session) {
  # Create a publisher client for comparison. This is not needed to create a viewer client.
  publisher_client <- connect()

  # Read the user-session-token header and create a viewer client
  user_session_token <- session$request$HTTP_POSIT_CONNECT_USER_SESSION_TOKEN
  viewer_client <- connect(token = user_session_token)

  output$default_api_key <- renderText(publisher_client$api_key)
  output$viewer_api_key <- renderText(viewer_client$api_key)
}

shinyApp(ui, server)
```

## Implementation notes:

- Call `get_oauth_credentials()` [here](https://github.com/posit-dev/connectapi/blob/c7d59b837543e573201d77c66b0c6fb6edc19506/R/get.R#L615-L628). It looks like it needs an additional field though: `"requested_token_type": "urn:posit:connect:api-key"`. Is that the field which makes this request special?

toph-allen:
## Intent

Add the ability to create a Connect client using content visitor-scoped authentication.

Fixes #362

## Approach

Added a token argument to the `connect()` function. When passed a token and running on Connect, performs an OAuth credential exchange request to get a visitor API key and returns a client using that API key instead. When an argument is provided to `token` but code is not running on Connect, prints a message and returns a client using the provided API key.

I considered a few tradeoffs:

- The general convention in R is to indicate optional arguments by giving them a default `NULL` value. This implementation breaks with the convention — `token` argument has no default value, and I use `missing()` to tell if it has been provided.
  - The default value for missing HTTP headers in R is `NULL`, so sticking to convention means that we can't tell the difference between the cases where a user just wants to use the API key, or is using the `token` arg but running the content locally.
  - This is actually _fine_ wrt functionality, because doing this leads to the correct behavior of, when `token` is provided correctly, using the API key locally but using the token on Connect.
  - _However_, I decided to break with the default `NULL` arg convention because this allows us to log a console message when the content is running locally and we see that they provided a NULL arg to `token` and are using the API key.
-  The `token` argument actually comes _before_ the pre-existing `prefix` argument. This is fine, I think: the `prefix` argument would be used when users want to set the env var value of both `server` and `api_key`, the first two arguments, so it should basically _never_ be called with them — which means it would _always_ be called by name and not positionally, and is fine to move. The `token` argument makes more sense to be up front with those two, although it's unlikely to ever be called positionally, because I doubt you'd ever call it and also pass in a `server` URL.
- The new logic was added in the `connect()` function, rather than the Connect object's `Connect$new()` initialization function. This… just made more sense to me. This is logic that wraps the direct initialization of the object.
  - Currently, the way that messaging happens inside `connect()`, when you use a token on Connect, the app logs will contain the line `Defining Connect with the server: {server}` twice. This is because that line is logged from `Connect$new()`. I guess we could move that log line out of that function, but that didn't seem like a huge priority. And technically, both clients _are_ defined, because the first one is used to make the OAuth request.
- I named the arg for the fallback API key `token_fallback_api_key` which is kinda verbose — I wanted to clarify that it wasn't, like, just a fallback API key that's used if the regular `api_key` is not present.
 
## Manual Verification

You can test this with the following Shiny app:

```r
library(shiny)
library(connectapi)

ui <- fluidPage(
  "Publisher",
  verbatimTextOutput("publisher_username"),
  verbatimTextOutput("publisher_api_key"),
  "Viewer",
  verbatimTextOutput("viewer_username"),
  verbatimTextOutput("viewer_api_key")
)

server <- function(input, output, session) {
  # Create a publisher client for comparison. This is not needed to create a viewer client.
  publisher_client <- connect()

  # Read the user-session-token header and create a viewer client
  user_session_token <- session$request$HTTP_POSIT_CONNECT_USER_SESSION_TOKEN
  viewer_client <- connect(token = user_session_token)

  output$publisher_username <- renderText(paste("username:", publisher_client$GET("v1/user")$username))
  output$publisher_api_key <- renderText(paste("api key:", publisher_client$api_key))
  output$viewer_username <- renderText(paste("username:", viewer_client$GET("v1/user")$username))
  output$viewer_api_key <- renderText(paste("api key:", viewer_client$api_key))
}

shinyApp(ui, server)
```

To publish this to Connect with the dev version of `connectapi`:

1. Place this app in a directory in a file named `app.R`.
2. In an R console, run `renv::init()`, `renv::install("posit-dev/connectapi@toph/viewer-api-key")`, and `renv::snapshot()`.
3. Publish to Connect.

## Checklist

### DIFF
--- a/R/connect.R
+++ b/R/connect.R
@@ -894,7 +894,7 @@ Connect <- R6::R6Class(
 #'
 #' When running on Connect, the client's environment will contain default
 #' `CONNECT_SERVER` and `CONNECT_API_KEY` variables. The API key's permissions
-#' will be scoped to the publishing user's account.
+#' are scoped to the publishing user's account.
 #'
 #' To create a client with permissions scoped to the content viewer's account,
 #' call `connect()` passing a user session token from content session headers
@@ -908,6 +908,10 @@ Connect <- R6::R6Class(
 #' @param token Optional. A user session token. When running on a Connect server,
 #'   creates a client using the content viewer's account. Running locally, the
 #'   created client uses the provided API key.
+#' @param token_fallback_api_key Optional. When a `token` is provided, but
+#'   content is running locally, this API key is used to create the client.
+#'   By default, the primary `api_key` is used, but by providing a different
+#'   key you can test a visitor client with differently-scoped permissions.
 #' @param prefix The prefix used to determine environment variables
 #' @param ... Additional arguments. Not used at present
 #' @param .check_is_fatal Whether to fail if "check" requests fail. Useful in
@@ -921,8 +925,15 @@ Connect <- R6::R6Class(
 #' \dontrun{
 #' client <- connect()
 #'
-#' # Running on Connect, create a client using the content visitor's account.
-#' client <- connect(token = session$request$HTTP_POSIT_CONNECT_USER_SESSION_TOKEN)
+#' # Running in Connect, create a client using the content visitor's account.
+#' # This example assumes code is being executed in a Shiny app's `server`
+#' # function with a `session` object available.
+#' token <- session$request$HTTP_POSIT_CONNECT_USER_SESSION_TOKEN
+#' client <- connect(token = token)
+#'
+#' # Use with a differently-scoped API key
+#' fallback_key <- Sys.getenv("VIEWER_SCOPED_API_KEY")
+#' client <- connect(token = token, token_fallback_api_key = fallback_key)
 #' }
 #'
 #' @examplesIf identical(Sys.getenv("IN_PKGDOWN"), "true")
@@ -935,8 +946,9 @@ Connect <- R6::R6Class(
 connect <- function(
     server = Sys.getenv(paste0(prefix, "_SERVER"), NA_character_),
     api_key = Sys.getenv(paste0(prefix, "_API_KEY"), NA_character_),
-    prefix = "CONNECT",
     token,
+    fallback_visitor_api_key = api_key,
+    prefix = "CONNECT",
     ...,
     .check_is_fatal = TRUE) {
   if (is.null(api_key) || is.na(api_key) || nchar(api_key) == 0) {
@@ -960,9 +972,10 @@ connect <- function(
       )
       con <- Connect$new(server = server, api_key = visitor_creds$access_token)
     } else {
+      con <- Connect$new(server = server, api_key = fallback_visitor_api_key)
       message(paste0(
         "Called with `token` but not running on Connect. ",
-        "Continuing with API key."
+        "Continuing with fallback API key."
       ))
     }
   }
--- a/R/connect.R
+++ b/R/connect.R
@@ -888,14 +888,26 @@ Connect <- R6::R6Class(
 
 #' Create a connection to Posit Connect
 #'
-#' Creates a connection to Posit Connect using the server URL and an api key.
+#' Creates a connection to Posit Connect using the server URL and an API key.
 #' Validates the connection and checks that the version of the server is
 #' compatible with the current version of the package.
+#'
+#' When running on Connect, the client's environment will contain default
+#' `CONNECT_SERVER` and `CONNECT_API_KEY` variables. The API key's permissions
+#' will be scoped to the publishing user's account.
+#'
+#' To create a client with permissions scoped to the content viewer's account,
+#' call `connect()` passing a user session token from content session headers
+#' to the `token` argument. To do this, you must first add a Connect API
+#' integration in your published content's Access sidebar.
 #'
 #' @param server The URL for accessing Posit Connect. Defaults to environment
 #'   variable CONNECT_SERVER
 #' @param api_key The API Key to authenticate to Posit Connect with. Defaults
 #'   to environment variable CONNECT_API_KEY
+#' @param token Optional. A user session token. When running on a Connect server,
+#'   creates a client using the content viewer's account. Running locally, the
+#'   created client uses the provided API key.
 #' @param prefix The prefix used to determine environment variables
 #' @param ... Additional arguments. Not used at present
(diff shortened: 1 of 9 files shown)


R/connect.R
```diff
@@ -954,6 +1001,8 @@ connect <- function(
   con
 }
 
+# viewer_client <- function(access_token)
```
mconflitti-pbc, line 1004: is this a leftover remnant of something else?

R/connect.R
```diff
@@ -933,6 +961,25 @@ connect <- function(
...
+    if (on_connect()) {
+      message()
+      visitor_creds <- get_oauth_credentials(
+        con,
+        user_session_token = token,
+        requested_token_type = "urn:posit:connect:api-key"
+      )
+      con <- Connect$new(server = server, api_key = visitor_creds$access_token)
+    } else {
+      con <- Connect$new(server = server, api_key = token_fallback_api_key)
```
mconflitti-pbc, line 975: May be worth recursively calling `connect()` here to ensure that the api_key provided runs through the checks defined on lines 954-960.

mconflitti-pbc (approved):
Couple suggestions but overall looks good

R/connect.R
toph-allen: Good point, I like that approach better!

mconflitti-pbc:
Only other comment is that the posit-sdk-py side of this is on hold until I can chat with Kelly about the naming convention she prefers us to use: Visitor vs Viewer across the sdks, docs, and connect server. So maybe this PR needs to be held up until then as well?

toph-allen:
> Only other comment is that the posit-sdk-py side of this is on hold until I can chat with Kelly about the naming convention she prefers us to use: Visitor vs Viewer across the sdks, docs, and connect server. So maybe this PR needs to be held up until then as well?

Sounds good, I'll hold off on merging till I hear from you about that. Thanks for moving that convo forward while I'm on jury duty!

R/connect.R
```diff
@@ -888,14 +888,30 @@ Connect <- R6::R6Class(
...
 #'   variable CONNECT_SERVER
 #' @param api_key The API Key to authenticate to Posit Connect with. Defaults
 #'   to environment variable CONNECT_API_KEY
+#' @param token Optional. A user session token. When running on a Connect server,
+#'   creates a client using the content visitor's account. Running locally, the
+#'   created client uses the provided API key.
+#' @param token_fallback_api_key Optional. When a `token` is provided, but
+#'   content is running locally, this API key is used to create the client.
+#'   By default, the primary `api_key` is used, but by providing a different
+#'   key you can test a visitor client with differently-scoped permissions.
```
mconflitti-pbc, line 914: Not sure what level of parity we want between the R and python packages but it was determined this wasn't necessary on the python side and on the publisher to handle when there is no user session token for now.

mconflitti-pbc: https://github.com/posit-dev/posit-sdk-py/pull/376#discussion_r1925713398

R/connect.R
```diff
@@ -933,6 +961,25 @@ connect <- function(
   }
   con <- Connect$new(server = server, api_key = api_key)
 
+  if (!missing(token)) {
+    error_if_less_than(con$version, "2025.01.0")
```
mconflitti-pbc, line 965: python is currently set to 2025.01.0-dev

R/connect.R
toph-allen: Ok, thanks! I'll take a look at the discussion there and probably follow suit

R/connect.R
toph-allen: `connectapi`'s version comparison mechanism strips off all suffixes and just looks at the digits — following our other precedents here :)

R/connect.R
toph-allen: Closing the loop on this, I chatted with @mconflitti-pbc some more. I think I'm going to leave the R code as-is, but rename the fallback key argument to `local_token_testing_key`.

Handling local testing automatically matches user expectations around R and Shiny, and handling local testing automatically raises the question of "What if the testing needs to user a different key?"

The difference between `connectapi` and `posit-sdk-py` here also feels to me like it might align with the differing ergonomics of the workflows between the packages (an argument to the client initializer in `connectapi` vs a method on an already-created client in `posit-sdk-py`). Like, in `connectapi`'s case, you're already calling the `connect()` function, so the thing to do when the token isn't present is obvious.

toph-allen: Marking this as resolved — thanks!

PR #367, opened 2025-01-29
feat: add `content_id` and `content_guid` fields to `get_jobs()`

toph-allen:
## Intent

Connect 2025.01.0 returns `content_id` and `content_guid` fields with `v1/content/{guid}/jobs` endpoints. `connectapi` has followed suit, adding those fields to data returned by `get_jobs()` and `get_job_list()`. The old values, `app_id` and `app_guid`, remain for now, but are deprecated and will be removed in a future update.

## Approach

- Nothing significant methodologically.
- Adjusted `compare_connect_version` and surrounding functions to allow it to be used consistently in conditional statements (previously I had made a change that allowed it to sometimes return `NA`; I see now that this just made things more complex).

## Checklist


R/content.R
```diff
@@ -111,7 +111,26 @@ Content <- R6::R6Class(
...
+      # for backward compatibility with older `connectapi` versions, which added
+      # a field named `app_guid` to job data.
+      if (compare_connect_version(self$connect$version, "2025.01.0") < 0) {
+        # Versions of Connect below 2025.01.0 only included `app_id`. We must
+        # add all other fields.
+        purrr::map(parsed, ~ purrr::list_modify(.x,
+          content_id = .x$app_id,
+          app_guid = self$content$guid,
+          content_guid = self$content_guid
+        ))
```
toph-allen, line 125: Although this logic would work by itself, I added the `if` statement here to make clearer what data we're expecting from the server in different conditions.

tdstein (approved):
Looks good. Straightforward.

PR #372, opened 2025-02-26
feat: Added product environment check helpers

### LINKED ISSUE #371: RSTUDIO_PRODUCT is deprecated; support POSIT_PRODUCT
This calls for helper(s) to determine which environment is being used. For example:
- `connectapi::is_running_on_connect()`
- `connectapi::is_running_on_workbench()`

or something similar to encapsulate the check.

Need to check both `RSTUDIO_PRODUCT` and `POSIT_PRODUCT` since older versions of connect and workbench will not have the latter set, but when it is set, they will be the same value. RSTUDIO_PRODUCT has not yet been removed, only deprecated for now.

Once merged, ensure that any documentation examples in our products use these helpers for the latest version of this utility. Add fallback note if not on version that includes these helpers.

mconflitti-pbc:
## Intent

Fixes #371 

## Approach

Adds environment check helpers to make it easier to determine where the code is running.

## Checklist

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -194,8 +194,28 @@ endpoint_does_not_exist <- function(res) {
     !("code" %in% names(httr::content(res, as = "parsed")))
 }
 
+get_product <- function() {
+  posit_product <- Sys.getenv("POSIT_PRODUCT")
+  if (posit_product != "") {
+    return(posit_product)
+  }
+  Sys.getenv("RSTUDIO_PRODUCT")
+}
+
+# Returns `TRUE` if we're running locally (no product env var set),
+# else `FALSE`.
+is_local <- function() {
+  get_product() == ""
+}
+
 # Returns `TRUE` if we're running on Connect as determined by the
-# `RSTUDIO_PRODUCT` env var, else `FALSE`.
+# `POSIT_PRODUCT` or `RSTUDIO_PRODUCT` env var, else `FALSE`.
 on_connect <- function() {
-  Sys.getenv("RSTUDIO_PRODUCT") == "CONNECT"
+  get_product() == "CONNECT"
+}
+
+# Returns `TRUE` if we're running on Workbench as determined by the
+# `POSIT_PRODUCT` or `RSTUDIO_PRODUCT` env var, else `FALSE`.
+on_workbench <- function() {
+  get_product() == "WORKBENCH"
 }
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -22,7 +22,7 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: local::., any::lintr, any::devtools, any::testthat
+          extra-packages: local::., any::lintr, any::devtools, any::testthat, any::cyclocomp
           needs: lint
 
       - name: Lint
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # connectapi (development version)
 
+## New features
+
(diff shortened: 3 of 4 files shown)


R/utils.R
```diff
@@ -194,8 +194,28 @@ endpoint_does_not_exist <- function(res) {
     !("code" %in% names(httr::content(res, as = "parsed")))
 }
 
+get_product <- function() {
```
toph-allen, line 197: Functions that we wanna export as part of the package namespace (i.e. loaded when running `library(connectapi)` should have `roxygen2` documentation, and that documentation must contain the `@export` tag. A good pretty light example to look at is the `get_runtimes()` documentation [here](https://github.com/posit-dev/connectapi/blob/f4008d92d023a9a6df03f8a2fb1a36afba3bd2c9/R/get.R#L689-L711). (Except for those docs contain a blank line, which — while it doesn't cause problems — is a mistake.)

`roxygen2` docs are preceded by `#'` and follow this general form (this is the block for `get_runtimes()`.

```r
#' Get available runtimes on server
#'
#' Get a table showing available versions of R, Python, Quarto, and Tensorflow
#' on the Connect server.
#'
#' @param client A `Connect` object.
#' @param runtimes Optional. A character vector of runtimes to include. Must be
#' some combination of `"r"`, `"python"`, `"quarto"`, and `"tensorflow"`. Quarto
#' is only supported on Connect >= 2021.08.0, and Tensorflow is only supported
#' on Connect >= 2024.03.0.
#' @return A tibble with columns for `runtime`, `version`, and `cluster_name`
#' and `image_name`. Cluster name and image name are only meaningful on Connect
#' instances running off-host execution.
#'
#' @examples
#' \dontrun{
#' library(connectapi)
#' client <- connect()
#' get_runtimes(client, runtimes = c("r", "python", "tensorflow"))
#' }
#'
#' @export
```

These won't need much. In fact, they could be listed on the same documentation page. To do that, you'd follow the pattern with `get_jobs()` (see [here](https://github.com/posit-dev/connectapi/blob/f4008d92d023a9a6df03f8a2fb1a36afba3bd2c9/R/content.R#L635-L714)). There, the return values for multiple content items are described i.e. (excerpt):

```r
#' @return
#'
#' - `get_jobs()`: A data frame with a row representing each job.
#' - `get_job_list()`: A list with each element representing a job.
```

and the end of the documentation includes the `@rdname` tag

```r
#' @rdname get_jobs
#' @export
```

Other functions can then include _just_ those same two lines as their documentation to show the same help page when queried (e.g. `get_job_list()` further down in the file).

That explanation is a little terse — let me know if you have any other questions!

toph-allen (changes requested):
The functions look good! Since they're public-facing functions, they need documentation and to be exported — see my comments for notes on that and hit me up if you have questions.

Not sure why linting is failing.

Also:

- After you write documentation you'll need to run `devtools::document()` to regenerate the documentation.
- You'll probably get CI failures from the `pkgdown` site not including some of the functions. The top-level `_pkgdown.yml` file needs entries for all exported functions. Some of its sections match e.g. all functions that start with `get_*` but the others will need to be listed manually. I learned this the hard way. 😂

toph-allen:
This is why CI is failing: `lintr` removed the dependency on `cyclocomp`: https://github.com/r-lib/lintr/releases/tag/v3.2.0

Changing line 25 of `.github/workflows/lint.yaml` to the following will probably fix:

```yaml
          extra-packages: local::., any::lintr, any::devtools, any::testthat, any::cyclocomp
```

toph-allen:
(update: you can pull main into this branch to fix linting)

PR #374, opened 2025-03-02
feat: add `get_packages()` and `get_content_packages()` functions

toph-allen:
## Intent

Add functions to wrap Connect's packages and content package dependencies APIs.

Not attached to a `connectapi` issue. However, this is in support of Connect documentation work (https://github.com/posit-dev/connect/issues/29920).

## Approach

Take the simplest possible approach, following existing patterns, for both the server-wide and content-specific package functions, with one exception:

- Right now the Packages API returns fields `app_id` and `app_guid`, but Connect is moving towards `content_id` and `content_guid` fields, so the `get_packages()` function only ever returns those newer names.

## Checklist


R/connect.R
```diff
@@ -824,6 +824,22 @@ Connect <- R6::R6Class(
       return(res[["schedules"]])
     },
 
+    # packages --------------------------------------------------
+
+    #' @description Get packages.
+    #' @param name The package name to filter by.
+    #' @param page_number The page number.
+    #' @param page_size Page size, max 500.
+    packages = function(name = NULL, page_number = 1, page_size = 500) {
```
jonkeane, line 833: We might have this pattern already, but do we want to keep on explicitly setting and passing all of these args to each endpoint that is paginated?

tests/testthat/test-get.R
```diff
@@ -193,3 +193,82 @@ test_that("get_vanity_urls() works", {
...
+          bundle_id = "6623",
+          app_id = "3652",
+          app_guid = "1935b6cb",
+          content_id = "3652",
+          content_guid = "1935b6cb"
+
+        )
+      )
+    )
+  )
```
jonkeane, line 259: Is this the same thing as the httptest response stuff up above? I'm a bit confused as to what the difference is here?

R/connect.R
toph-allen: Yeah, the arguments exposed by paginated methods and functions are widespread and consistent in the package.

The way that pagination functions are implemented in `connectapi`, the endpoints have to explicitly receive page number and size arguments (or at least page number), because pagination is implemented by using `rlang` to create a copy of the quosure, increment the page number in the expression, and call the resulting new quosure ([ref](https://github.com/posit-dev/connectapi/blob/c3ef91ed95e6a4da6bb2f3f868bd6eeefebb6f66/R/page.R#L15-L38)).

As far as the user facing side (`get_packages()`) — I think with paginated endpoints as currently implemented, having controls over the page size and limit are important for development, because, as with all our paginated endpoints, the `packages` endpoint is limited to returning 500 items for pages, so for example calling it on Dogfood takes 7 minutes (I tested in `posit-sdk-py`). So when you're developing with it and iterating, it's nice to be able to cut it off after a certain point.

```
In [14]: timeit(
    ...:     lambda: pd.DataFrame(client.packages.fetch()),
    ...:     number=1
    ...: )
Out[14]: 437.47203699999955
```

tests/testthat/test-get.R
toph-allen: P.S. If you think it's better to use the same mocking approach in each case — you're probably right, and I'll move the `httptest` fixtures into this file.

jonkeane: > If you think it's better to use the same mocking approach in each case — you're probably right, and I'll move the httptest fixtures into this file.

Yeah, I find the cognitive overhead of keeping track of two different mocking systems (+ the inherent overhead of mocking itself) makes understanding things really hard. Could we pick one and use that one?

R/connect.R
toph-allen: Discussed this with @jonkeane and I want to acknowledge that conversation here: they think that including access to pagination controls for end users is the wrong choice.

I conditionally disagree — I agree that it'd be much better to hide them, but feel strongly that as long as Connect's endpoints are paginated in a way that makes them deliver data very slowly, we need to include controls like `limit` to make developing anything with these endpoints at all possible.

toph-allen: Discussed this with @jonkeane and I want to acknowledge that conversation here: they feel strongly that including access to pagination controls for end users is the wrong choice.

I conditionally disagree — I agree that it'd be much better to hide them, but feel strongly that as long as Connect's endpoints are paginated in a way that makes them deliver data very slowly, we should include controls like `limit` to make developing anything with these endpoints at all possible.

For now, I'm going to merge this with the controls in place — I appreciate you letting me trust my gut and experience on this.

Removing pagination controls should be a priority for APIs where… pagination isn't problematic.

toph-allen:
Merging after discussion with @jonkeane a few days ago. It sounds like they assent to merging while disagreeing on an aspect of the approach.

PR #375, opened 2025-03-03
chore: fix linting

toph-allen:
`lintr` v3.2.0 removed a hard dependency on `cyclocomp`, so the cyclocomp linter is failing. This PR installs that package to the lintr CI workflow, and should fix the problem.


toph-allen:
`lintr` 3.2.0 includes a new default `return_linter()` which, by default, enforces tidyverse return style. This is the style we prefer, but for now I'm disabling this linter as there's a lot of old code that uses explicit returns.

PR #378, opened 2025-03-11
fix: simplify `get_packages()`

toph-allen:
## Intent

Simplify `get_packages()`, removing pagination-related parameters.

Fixes #

## Approach

- Internally, use a response limit of 100000 items for the endpoint.
- Remove `page_size` and `limit` from the user-facing function `get_packages()`.
- To fix tests that broke, alter the `page_offset()` function to cease iteration when the total number of aggregated items reaches the total number of items that the server will return. (Previously, it would continue to request a new page until the number of new items was 0, but the responses always contain the total number of items to be returned!)

## Checklist

### DIFF
--- a/R/connect.R
+++ b/R/connect.R
@@ -830,12 +830,12 @@ Connect <- R6::R6Class(
     #' @param name The package name to filter by.
     #' @param page_number The page number.
     #' @param page_size Page size, max 500.
-    packages = function(name = NULL, page_number = 1, page_size = 500) {
+    packages = function(name = NULL, page_number = 1) {
       url <- v1_url("packages")
       query_params <- list(
         name = name,
         page_number = page_number,
-        page_size = page_size
+        page_size = "1000000"
       )
       self$GET(url, query = query_params)
     },
--- a/R/connect.R
+++ b/R/connect.R
@@ -826,16 +826,16 @@ Connect <- R6::R6Class(
 
     # packages --------------------------------------------------
 
-    #' @description Get packages.
+    #' @description Get packages. This endpoint is paginated.
     #' @param name The package name to filter by.
-    #' @param page_number The page number.
-    #' @param page_size Page size, max 500.
-    packages = function(name = NULL, page_number = 1) {
+    #' @param page_number Page number.
+    #' @param page_size Page size, default 100000.
+    packages = function(name = NULL, page_number = 1, page_size = 100000) {
       url <- v1_url("packages")
       query_params <- list(
         name = name,
-        page_number = page_number,
-        page_size = "1000000"
+        page_number = format(page_number, scientific = FALSE),
+        page_size = format(page_size, scientific = FALSE)
       )
       self$GET(url, query = query_params)
     },
--- a/R/connect.R
+++ b/R/connect.R
@@ -834,6 +834,9 @@ Connect <- R6::R6Class(
       url <- v1_url("packages")
       query_params <- list(
         name = name,
+        # We use `format()` here because `httr` seems to serialize integers over
+        # 99999 with scientific notation, which causes incorrect query strings.
+        # Finding a more general solution to this is tracked by #379.
         page_number = format(page_number, scientific = FALSE),
         page_size = format(page_size, scientific = FALSE)
       )
--- a/R/get.R
+++ b/R/get.R
@@ -747,9 +747,15 @@ get_runtimes <- function(client, runtimes = NULL) {
 #' @description Get a data frame of package dependencies used by all content
 #' items on the server.
 #'
+#' @usage get_packages(src, name = NULL, page_size = 100000, limit = Inf)
+#'
 #' @param src A `Connect` client object.
 #' @param name Optional package name to filter by. Python package are normalized
 #' during matching; R package names must match exactly.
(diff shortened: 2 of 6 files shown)


R/connect.R
```diff
@@ -830,12 +830,12 @@ Connect <- R6::R6Class(
...
     #' @param page_number The page number.
     #' @param page_size Page size, max 500.
-    packages = function(name = NULL, page_number = 1, page_size = 500) {
+    packages = function(name = NULL, page_number = 1) {
       url <- v1_url("packages")
       query_params <- list(
         name = name,
         page_number = page_number,
-        page_size = page_size
+        page_size = "1000000"
```
jonkeane, line 838: Should this really be a string?

R/connect.R
```diff
@@ -830,12 +830,12 @@ Connect <- R6::R6Class(
     #' @param name The package name to filter by.
     #' @param page_number The page number.
     #' @param page_size Page size, max 500.
-    packages = function(name = NULL, page_number = 1, page_size = 500) {
+    packages = function(name = NULL, page_number = 1) {
```
jonkeane, line 833: I'm not opposed to having this as an argument if it defaults to something sensible. In case someone really does know what they're doing and wants to fiddle with it that's ok. My complain in prior PRs was that we _force_ folks to deal with this in cases where they don't want to. 

But if you would rather hide it entirely, that's fine.

R/connect.R
```diff
@@ -830,12 +830,12 @@ Connect <- R6::R6Class(
     #' @param name The package name to filter by.
     #' @param page_number The page number.
     #' @param page_size Page size, max 500.
```
jonkeane, line 832: This will need updating / correcting

tests/testthat/test-get.R
```diff
@@ -226,7 +226,7 @@ test_that("get_packages() works as expected with current return value", {
     )
   )
   expect_identical(
-    get_packages(client, page_size = 2, limit = 2),
+    get_packages(client),
```
jonkeane, line 229: What was `limit = 2` here before? I don't see that being removed, but maybe it was just eaten in `...` somewhere?

Regardless, this is one place where having the ability to specify makes testing easier (in a way that encourages good factoring of our own code...). This particular parameter is just passed through so probably NBD to have a test that confirms that, but if there was any sort of processing or whatever it certainly would make something like that easier.

jonkeane (changes requested):
Thanks for this

R/connect.R
tdstein, line 838: I arbitrarily picked the 1M value in Python since `requests` has no response body limit. Do you think the same value makes sense here?

IIRC, dogfood is somewhere in the 100k range. Since we haven't been able to test beyond that number, should we use 100k instead? That'll result in a few round trips on dogfood, but at least we will know things don't blow up.

toph-allen: Yeah, 100k or 500k seems reasonable.

R/connect.R
toph-allen: Thanks

R/connect.R
toph-allen: You're right, this should be handled at a different layer.

tests/testthat/test-get.R
toph-allen: [This](https://github.com/posit-dev/connectapi/pull/378/files#r1992153156) is where `limit` was removed from.

---

Side note, slightly related: `connectapi` used to paginate until the API returned empty pages, which required explicitly setting the limit at 2 in the tests. The pagination quirk has now been fixed: paginating until the response is empty is not required because the responses provide the `total` number of objects to be returned.

R/connect.R
toph-allen: fwiw, I don't _think_ `httr` has a body response limit either

toph-allen: Holding myself back from trying to do this for all query params in all requests for now

R/get.R
```diff
@@ -781,16 +772,14 @@ get_runtimes <- function(client, runtimes = NULL) {
 #'
 #' @family packages functions
 #' @export
-get_packages <- function(src, name = NULL, page_size = 500, limit = Inf) {
+get_packages <- function(src, name = NULL) {
```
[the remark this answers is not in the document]
jonkeane: Aaah I see now, thanks

R/page.R
```diff
@@ -69,10 +69,11 @@ page_offset <- function(client, req, limit = Inf) {
 
   req_response <- rlang::eval_tidy(qreq)
   res <- req_response$results
+  total_items <- req_response$total
 
   agg_response <- res
   agg_length <- length(agg_response)
-  while (length(res) > 0 && agg_length < limit) {
+  while (length(res) > 0 && agg_length < limit && agg_length < total_items) {
```
tdstein, line 76: Is limit every greater than total_items? Just curious what sparked this change.

tdstein (approved):
I'm good with this. For future reference, it's worth updating the pull request description to reflect changes made during the review process.

toph-allen:
@tdstein 🫡 Ah yeah, thanks for pointing that out

jonkeane (approved):

PR #386, opened 2025-03-24
set scipen in request method

### LINKED ISSUE #379: integer query parameters are serialized with scientific notation when over 99999
When making requests, query parameters that are integers (perhaps also doubles) greater than `99999` are serialized as scientific notation unless explicitly converted with `format(x, scientific = FALSE)`.

It seems that base `httr` functions like this; it does not seem like this is an artifact of the way we wrap it.

```r
> httr::GET("http://google.com", query = list(number = 100000))
Response [[http://www.google.com/?number=1e%2B05]](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)
```

For now, there is a kludgey solution in the `Connect$packages()` method, but there should be a better solution, probably in `Connect$request()`.

https://github.com/posit-dev/connectapi/blob/67a52e20e147a9f1102ac73c73370c7498913a47/R/connect.R#L837-L841

toph-allen:
## Intent

When large query parameters are serialized, they use R's default serialization rules for scientific notation. This PR fixes that, preventing scientific notation for request query params (shorter than 999 digits).

Fixes #379 

## Approach

- Set the option `scipen` to `999` locally in the `Connect$request` method.
- Add a test which both confirms correct serialization, and that the value is unchanged in the calling scope.

## Checklist

- [ ] Does this change update `NEWS.md` (referencing the connected issue if necessary)? **Not needed**
- [ ] Does this change need documentation? Have you run `devtools::document()`? **Not needed**


nealrichardson (approved):
Nice clean fix! Looks like you need to add a newline at the end of the test file to appease the linter, but otherwise you're good to go!

R/connect.R
```diff
@@ -132,6 +132,9 @@ Connect <- R6::R6Class(
     #' `httr::content(res, as = parser)`.
     #' @param ... Additional arguments passed to the request function
     request = function(method, url, ..., parser = "parsed") {
+      old_opt <- options(scipen = 999)
+      on.exit(options(old_opt), add = TRUE)
```
jonkeane, line 136: 💯 Even got the various slightly funky things around `options()` (that it returns its former self!) and `add` with `on.exit()` (why that is not default I will never know!)

Issue #384, opened 2025-03-18
Assume Viewer Credentials when running on Connect


toph-allen:
The general expectation when running applications on Connect is that the application is running using the visitor's credentials. This is not the current behavior. Instead, `connectapi` uses the `CONNECT_API_KEY` variable, which belongs to the content owner, not the visitor.

Therefore, the default behavior should be as follows:

When running on Connect, assume that a visitor integration is configured. If the integration is not set, provide a sensible error message that directs the user to create the integration or explicitly overrides this behavior. When the integration is set, it assumes the visitor's credentials via the HTTP header.

When running off Connect, assume the current behavior (e.g., `CONNECT_API_KEY` is set).

(Sibling issue to https://github.com/posit-dev/posit-sdk-py/issues/398)

toph-allen:
In `connectapi`, right now, we rely on the user to explicitly pass in the visitor token from the Shiny session object. Getting this automatically will require making some assumptions about the environment we're running in, I think. I'll need to think through this some more.

toph-allen:
Random thought: perhaps this will need to be implemented as "When `connect()` is called within a Shiny `server()` function, and is running on Connect, it uses the visitor API key".

toph-allen:
Proposals to simplify usage and make visitor auth feel more "default" on Connect.

tl;dr: If the `connect()` client constructor function takes a Shiny session object with the default argument `getDefaultReactiveDomain()`, it can automatically use visitor credentials when called within a Shiny server function.

### As currently implemented

Requires that the user explicitly pass in the token header. Even *I*, the person who wrote it, don't remember what the header is called.

```r
library(shiny)
library(connectapi)
library(glue)

ui <- fluidPage(
  textOutput("who_am_i")
)

server <- function(input, output, session) {

  # Visitor authenticated client when running on Connect
  token <- session$request$HTTP_POSIT_CONNECT_USER_SESSION_TOKEN
  client <- connect(token = token)

  me <- client$me()
  output$who_am_i <- renderText(
    glue("Authenticated as {me$first_name} {me$last_name} ({me$email})")
  )
}

shinyApp(ui, server)
```

### Proposal: Automatically get the session when in a reactive context

In Shiny, there's already a pattern of function, such as `updateSelectInput()`, which take `session` as their first arg with a default value of `getDefaultReactiveDomain()`. Experimentally, `getDefaultReactiveDomain()` can get the token whether it's called inside a reactive context or not.

[...]

PR #391, opened 2025-04-15
feat: use viewer credentials by default if possible

toph-allen:
## Intent

Automatically assume the viewer's credentials on Connect if possible.

~Fixes~ Related #384 

## Approach

As of 2025-04-15, the current draft includes a lot of conditional logic. It flows kind of confusingly right now, and I want to simplify that, but first I want some input on whether I'm thinking this through correctly.

There are two test apps available. The first uses the released version of `connectapi`, the second uses this dev version:

https://rsc.radixu.com/default-viewer-credential-test-0-7-0/
cleanshot_2025-04-15_at_15 21 44

https://rsc.radixu.com/default-viewer-credential-test-devel/
cleanshot_2025-04-15_at_15 21 51

The function `connect()` can be called with an API key or token passed explicitly, or with nothing provided.

- Nothing provided (just called as `connect()`):
  - Locally, use the credentials associated with the API key in the environment var
  - On Connect, use viewer credentials if possible*, otherwise use the credentials from the environment var API key.
- API key explicitly provided:
  - Locally, use the credentials associated with the API key
  - On Connect, use the credentials associated with the API key
- Token explicitly provided:
  - Locally, use the API key's credentials (or a testing key's credentials)
  - On Connect, use the publisher's API key to get the token's credentials

\* This is only possible in Shiny apps, and on Connect v2025.01.0 or later. This adds more conditional branches to emit the correct messaging.

I'm going to take a second look at the code and different conditions we account for.

## Checklist

### DIFF
--- a/R/connect.R
+++ b/R/connect.R
@@ -1017,13 +1017,13 @@ connect <- function(
         user_session_token = token,
         requested_token_type = "urn:posit:connect:api-key"
       )
-      con <- connect(server = server, api_key = visitor_creds$access_token)
+      con <- Connect$new(server = server, api_key = visitor_creds$access_token)
     } else {
       message(paste0(
         "Called with `token` but not running on Connect. ",
         "Continuing with fallback API key."
       ))
-      con <- connect(server = server, api_key = token_local_testing_key)
+      con <- Connect$new(server = server, api_key = token_local_testing_key)
     }
   }
 
--- a/R/connect.R
+++ b/R/connect.R
@@ -981,8 +981,36 @@ connect <- function(
   }
   con <- Connect$new(server = server, api_key = api_key)
 
+  if (on_connect()) {
+    comp <- compare_connect_version(using_version, "2025.01.0")
+    if (comp < 0) {
+      if (!missing(token)) {
+        # If running on a too-old version of Connect and token was explicitly
+        # passed, error.
+        stop(
+          "Running content using the viewer's credentials ",
+          "requires Connect v2025.01.0 or later."
+        )
+      } else {
+        # If running on too-old Connect and no token was explicitly passed, use
+        # the old behavior (run with publisher creds).
+        message(
+          "Content will run using publisher's credentials. ",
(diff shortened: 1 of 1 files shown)


R/connect.R
```diff
@@ -981,21 +988,53 @@ connect <- function(
...
   if (!missing(token)) {
-    error_if_less_than(con$version, "2025.01.0")
     if (on_connect()) {
       visitor_creds <- get_oauth_credentials(
         con,
         user_session_token = token,
         requested_token_type = "urn:posit:connect:api-key"
       )
-      con <- connect(server = server, api_key = visitor_creds$access_token)
+      con <- Connect$new(server = server, api_key = visitor_creds$access_token)
```
mconflitti-pbc, line 1031: This will fail unless the integration is attached to the content. I dont remember where that landed but will you be using the viewer api key integration or something in the rstudio-credentials header?

toph-allen: The viewer API key for now, as unless I'm mistaken we haven't moved that work forward yet.

I'm aware that it'll fail on first deploy in that case, and (cc @jonkeane) I think we're willing to accept that for now? I do think better error messaging in `get_oauth_credentials()` in this function are critical to smoothing that experience over! `get_oauth_credentials()` currently just prints a `400` message, and that should be improved (in a different PR). This function should catch that error and add to the message to specifically advise on the type of integration to select.

@jonkeane I guess whether to proceed with landing this PR now hinges on whether we think "initial deploy fails to run until you add the integration" or "runs with admin credentials" is the least bad user experience. Thoughts?

jonkeane: Failing on initial deploy until configuration is correct is (unfortunately) pretty common for Connect today, so this isn't _much worse_ than other cases (e.g. when someone needs to set a specific env var). But it *is* a poor experience, and yet another toe-stub on not having the visitor API key on automatically.

toph-allen:
Closing as this approach is no longer how we're going to solve this problem.

Issue #328, opened 2024-11-08
remove workaround for `v1/groups` endpoint


toph-allen:
The `get_groups()` function has a workaround for a pagination quirk in Connect's `v1/groups` API, which always returns the first page of results when a `prefix` is specified.

This bug is tracked in the Connect repo: https://github.com/rstudio/connect/issues/28774

When that's resolved, we should remove the workaround: https://github.com/rstudio/connectapi/pull/326#discussion_r1834423806

nealrichardson:
The groups API will support pagination in 2025.04.0, but we don't want to just revert #326 because we'll still need that for compatibility with older versions. So we'll want to do things differently based on the Connect server version.

PR #392, opened 2025-04-25
chore: format everything with air

nealrichardson:
## Intent

I started working on #328 and found that `air` wanted to rewrite a lot of other stuff whenever I saved a file, so let's just turn everything over to the 🤖 s all at once.

## Approach

```
air format .
git add .
git commit
git push
```

Then, some refactoring:

* Looks like something is tighter in R 4.5 that is causing the `set_environment_remove()` function to error, unrelated to the formatting--it's failing on other PRs too. I'm not sure how it ever worked but 🤷 . I added some new non-integration tests, and then fixed the offending call. (Created #394 to generally do better here.)
* lintr and air disagree about how to indent long `if` statements, so I did some refactoring to avoid the issue.
* There was also one `nolint` comment that got misplaced in the formatting

Review those fixes in https://github.com/posit-dev/connectapi/compare/ffa11f870bc75a955d884365f2aa9ab65a9f4771...00895f5392e764a328fb560525dd154a0b7013b9


nealrichardson:
I don't understand the integration test failure, or really, I don't understand how that code worked at all before. Maybe something got more strict in the R 4.5 release and this is the first time we're running integration tests after?

As for linting, I think we need to switch to using air to lint too?

toph-allen (approved):
Today I learned that `air` exists!

This looks great — thanks @nealrichardson. I'm good to merge as-is. Agreed with your point that we should switch to `air` for linting (if it also does linting, I don't see that mentioned on its site). If not, perhaps there are `lintr` rules we could change to prevent disagreement?

nealrichardson:
> Today I learned that `air` exists!
> 
> This looks great — thanks @nealrichardson. I'm good to merge as-is. Agreed with your point that we should switch to `air` for linting (if it also does linting, I don't see that mentioned on its site). If not, perhaps there are `lintr` rules we could change to prevent disagreement?

It doesn't do linting, but you can run it in a mode where it just reports the changes and doesn't fix, I think? At least I think that's what we switched to doing in the Connect repo. 

Disagreements were minor, and really they were places where by refactoring the code we made it more readable anyway. So let's see how it goes and we can re-evaluate later.

PR #393, opened 2025-04-25
feat: update `get_groups()` to support pagination, if Connect version supports

### LINKED ISSUE #328: remove workaround for `v1/groups` endpoint

nealrichardson:
## Intent

Fixes #328. Based on #392, will rebase after that lands.

## Approach

On old Connect, set the `page_size` and `limit` to whatever it takes to satisfy the requested `limit`, up to 500, the max the API allows. That way, we'll get everything (to the best of our ability) on one page, and won't request additional pages, which the server did not support (it would ignore page_number and keep returning page 1, which would make the paginating client here just cycle forever). 

On new Connect, just paginate. 

Added tests for both cases, both of the `get_groups()` function that wraps the pagination, and also some `expect_GET`s thrown in there to be explicit about what requests are being made. 

## Checklist


nealrichardson:
This is rebased and ready for review.

R/groups.R
```diff
@@ -4,10 +4,11 @@
...
 #' The filter is case insensitive.
-#' @param limit The number of groups to retrieve before paging stops.
-#'
-#' `limit` will be ignored is `prefix` is not `NULL`.
-#' To limit results when `prefix` is not `NULL`, change `page_size`.
+#' @param limit The number of groups to retrieve before paging stops. Default
+#' is to return all results; however, for Connect server versions older than
+#' 2025.04.0, `limit` is capped at 500 when `prefix` is provided. If your search
+#' prefix would return more than 500 groups, try using multiple queries with
+#' more selective `prefix` values and append the results.
```
toph-allen, line 11: I think we can probably strike the last sentence — my initial reaction is that the added context is helpful, but the advice only applies to older Connects, and that isn't _entirely_ clear, which makes it a little confusing.

(Other points in favor of striking it: (1) it's probably a significant edge case; I don't _think_ we've had any customer reports of hitting it, (2) if a customer really needs to figure out how to access more groups with a prefix, they can probably figure it out.)

Actually — would better advice not just be to get_groups without specifying `prefix` and filter within R? Probably.

toph-allen (approved):
Looks good to me — thanks for circling back on this!

R/groups.R
nealrichardson: I removed the last sentence, all good? I agree it's not worth saying much on here, it is a real edge case to have that many groups I'd think.

Issue #308, opened 2024-10-07
chore: update integration tests to 2024.09.0


toph-allen:
It looks like content used in the integration tests uses versions of R that are no longer in the Docker images (e.g. `tests/testthat/examples/shiny` uses R 3.6.3).

jonkeane:
I see that you rolled back something in your other PR — mind adding some details about what needs updating here?

toph-allen:
@jonkeane done.

PR #396, opened 2025-04-28
ci: add to the integration testing matrix

### LINKED ISSUE #395: chore: update integration testing matrix
The latest version in https://github.com/posit-dev/connectapi/blob/main/.github/workflows/integration-tests.yaml is a year old.

nealrichardson:
## Intent

Fixes #395

## Approach

Add to the matrix

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -15,6 +15,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - "2025.03.0" # jammy
+          - "2024.09.0" # jammy
           - "2024.03.0" # jammy
           - "2023.09.0" # jammy
           - "2023.03.0" # bionic


nealrichardson:
@toph-allen I see #308, is that why these are failing? It's hard to tell--I made #397 to make the tests quieter so it's easier to tell what is failing and what is just making noise.

nealrichardson:
> Basically (I think), the manifests for the test content specify R 3.6.3, which is not present on the Docker images, so Connect fails to build those content items. To get things working, you'll need to spin up `renv` environments in the test content to generate new manifests with newer R versions.
> 
> Those new R versions would also need to be present on the older images, too, I think (because we only have one manifest per content item for the entire matrix), so you may wind up needing to remove some old Connect versions from the matrix.
> 
> Let me know if I didn't explain any of that clearly!

Clearly explained! But 🤮 ! Let's see if we can find a better way. The apps/reports are simple themselves, maybe we can generate manifests on the fly somehow. 

Alternatively, we could refactor the tests so that we only use static/dependency-free content bundles--we aren't testing that Connect runs apps correctly here, we're just testing the API integration.

toph-allen:
> Clearly explained! But 🤮 ! Let's see if we can find a better way. The apps/reports are simple themselves, maybe we can generate manifests on the fly somehow.
> 
> Alternatively, we could refactor the tests so that we only use static/dependency-free content bundles--we aren't testing that Connect runs apps correctly here, we're just testing the API integration.

Some of the failing content powers things like parameterized R Markdown / variant APIs, so while I like the second option better for simplicity's sake, it wouldn't be possible to keep the current suite of integration tests with *just* static content. For example:

https://github.com/posit-dev/connectapi/blob/4a4a8107d7ec7d46c2fb61f301336a48add1974e/tests/integrated/test-content.R#L431-L455

Looks like the setup code for those tests is [here]( [setup code](https://github.com/posit-dev/connectapi/blob/4a4a8107d7ec7d46c2fb61f301336a48add1974e/tests/integrated/test-content.R#L1-L123)). To generate manifests on the fly, we could use `rsconnect::deployApp()` etc., which generate already do the manifest-generation on the fly. Or actually, perhaps `rsconnect::writeManifest()` would be a simpler drop-in replacement, that would still let most of the existing code run untouched.

nealrichardson:
> To generate manifests on the fly, we could use `rsconnect::deployApp()` etc., which generate already do the manifest-generation on the fly. Or actually, perhaps `rsconnect::writeManifest()` would be a simpler drop-in replacement, that would still let most of the existing code run untouched.

I guess we could do that in the test suite if we can run the tests using a version of R that is supported in all of the Connect images in our matrix. That should be doable? Or we need to expand the matrix definition to specify a tuple `(connect_version, r_version)`. 

2022.08.0 had R 4.1 https://github.com/rstudio/rstudio-docker-products/commit/cfd6a875d1f88ee1e081a68e978b12819bc4ddbd#diff-848e5afd960438beae1c9fdce3527cb65495623dd1a2eece90289719871d56a8R64

And so does 2025.04.0 from yesterday: https://github.com/rstudio/rstudio-docker-products/blob/main/connect/Dockerfile.ubuntu2204#L7

So maybe we target 4.1 with the integration tests, and drop any older Connect versions that don't have R 4.1? 2022.08 is beyond the support window anyway.

nealrichardson:
Closing in favor of #417.

PR #399, opened 2025-04-29
feat: include vanity urls in get_content

### LINKED ISSUE #398: fix: include vanity url in `get_content()` results
`get_content()`'s documentation states that the content URL either contains the content URL or the vanity URL for a content item:

https://github.com/posit-dev/connectapi/blob/02acf5e248b5d40300b51ccadf863147e6a19a6d/R/get.R#L176-L177

This is not the case; it only fetches the regular "open solo" URL. It should function as documented.

toph-allen:
## Intent

Include vanity URLs in `get_content()` data frame.

Fixes #398 

## Approach

- Include vanity URLs in `get_content()` results.
- Update `get_content()` documentation to detail all returned fields.

I considered a slightly more complex approach, bubbling up `include` to the top level and allowing users to customize the returned data, but the performance penalty for including all the fields seems negligible, so I don't think it's worth the time and complexity, and decided to keep things as simple as possible.

N.B. Integration tests haven't finished yet, this might require a test change…

## Checklist


R/get.R
```diff
@@ -203,6 +220,7 @@ get_content <- function(
   guid = NULL,
   owner_guid = NULL,
   name = NULL,
+  include = c("tags", "owner", "vanity_url"),
```
nealrichardson, line 223: I know vanity_url was added relatively recently to the API here. How should this behave for older versions of Connect where it's not available here?

toph-allen: Oh, I didn't realize that it was new! I think this PR isn't ready for prime time — I'll mark it as a draft and ping you when it's ready, after I've done some experimentation.

nealrichardson: Looks like it was added in 2024.06.0: https://docs.posit.co/connect/news/#posit-connect-2024.06.0-new

So not new new, but well within the support window.

toph-allen: Thanks. Added version-keyed behavior in the latest push, plus tests to make sure the correct requests are sent to different versions (the endpoint returns a `400` if an invalid parameter is supplied)

tests/testthat/test-get.R
```diff
@@ -360,3 +360,39 @@ test_that("get_packages() works as expected with `content_guid` names in API res
     )
   )
 })
+
+test_that("get_content only requests vanity URLs for Connect 2024.06.0 and up", {
+  with_mock_dir("2024.05.0", {
+    client <- Connect$new(server = "http://connect.example", api_key = "not-a-key")
+    client$version
```
toph-allen, line 367: `version` is fetched lazily, so calling it here just ensure that it's fetched from the correct mock dir before we `expect_GET` within `without_internet()`.

toph-allen:
@nealrichardson Now has Connect-version-dependent behavior.

tests/testthat/test-get.R
nealrichardson: How about this, to memorialize that comment and also be explicit:

```suggestion
    # Call $version here because it is fetched lazily, and we need it populated
    # before we do `without_internet()`
    expect_equal(client$version, "2024.05.0")
```

If you agree, you'll want to do the same for the other two tests

tests/testthat/test-get.R
```diff
@@ -327,3 +327,39 @@ test_that("get_packages() works as expected with `content_guid` names in API res
...
+  })
+
+  with_mock_dir("2024.07.0", {
+    client <- Connect$new(server = "http://connect.example", api_key = "not-a-key")
+    client$version
+  })
+  without_internet({
+    expect_GET(
+      get_content(client),
+      "http://connect.example/__api__/v1/content?include=tags%2Cowner%2Cvanity_url"
```
nealrichardson, line 361: I'm guessing that you'll need to update the other mocks for `GET v1/content/`, at least the mock file paths to expect this query string.

R/get.R
```diff
@@ -208,7 +225,22 @@ get_content <- function(
 ) {
   validate_R6_class(src, "Connect")
 
-  res <- src$content(guid = guid, owner_guid = owner_guid, name = name)
+  if (compare_connect_version(src$version, "2024.06.0") < 0) {
```
nealrichardson, line 228: Probably not necessary but I'd add a comment here explaining that vanity_url was only added to this API in 2024.06.0, just to be explicit about why we're doing this.

nealrichardson (commented):
This looks like a good direction!

R/get.R
toph-allen: Done. For me, this is right on the line between "worth documenting" and "the code itself explains what the difference is" — I'd be fine either way.

tests/testthat/test-get.R
toph-allen: Yeah — if it was worth a comment for the reviewer, probably worth a comment for future editors of the code. :)

tests/testthat/test-get.R
toph-allen: None of the other mocks seem to have broken, actually — no failed tests, just an angry linter.

R/get.R
nealrichardson: Yeah I hear you, it's on the line. I personally lean towards over-commenting, I've never regretted having left extra explanatory comments in code the next time I come back to it, but definitely have wished I'd left more back when I understood what I was doing.

tests/testthat/test-get.R
nealrichardson: That's odd, do we not have mock tests for `get_content()`? I would expect something to fail, unless we're only testing on the old API and were always including tags and owner before?

toph-allen: I see what you mean.

Looking at this more in more depth: the default mock path is named `"2024.08.0"` but actually runs with an empty server version (which is a legitimate Connect configuration, but a poor name for that directory.

I've been mulling over in the back of my head how to address this — the sheer number of files in the mock directory makes it kind of difficult to grapple with in any way that, like, changes the hashes of many requests at the same time.

I think in the short term, renaming the default mock path to `"unversioned"` or `"null_version"` would make it more obvious what's going on. I don't know if that addresses the underlying problem or not, though — I feel like there _is_ an underlying problem with the organization of mocks (that's worth writing up in an issue / tackling separately).

nealrichardson: One idea is remove the default mock path (https://github.com/posit-dev/connectapi/blob/main/tests/testthat/setup.R#L36-L38) and always use `with_mock_dir()` explicitly pointing at a path. And/or yes, having a dir that is for the case where the API suppresses the server version could be useful for something I guess.

nealrichardson: To my original question: the default for `include` in `connect$content()` is `tags,owner`, so on the no-version-exposed API, the request is the same as it was before.

nealrichardson (approved):
Looks good. I think there are some followups to create:

* Sounds like you have something in mind about the mock directories, per the discussion in the comment.
* It would be good to take a pass through the other functions that do stuff with vanity URLs and see if they can be simplified at least on new Connect. `get_vanity_url`, for example, may already have the vanity URL and not need to make an additional request.

toph-allen:
@nealrichardson Thanks, I'll file those

tests/testthat/test-get.R
toph-allen: > One idea is remove the default mock path ([`main`/tests/testthat/setup.R#L36-L38](https://github.com/posit-dev/connectapi/blob/main/tests/testthat/setup.R?rgh-link-date=2025-04-30T21%3A13%3A26Z#L36-L38)) and always use `with_mock_dir()` explicitly pointing at a path. And/or yes, having a dir that is for the case where the API suppresses the server version could be useful for something I guess.

Yeah, I think this might be the case, at least going forward. Even code that doesn't explicitly check a version of Connect is tested against mocks captured from a specific version, and good practice might be to just use that mock dir.

And yep, your answer to your original question is correct.

PR #400, opened 2025-04-30
fix: minor improvements to time zone parsing

toph-allen:
## Intent

#291 fixed a bug in time zone parsing. This simplifies the approach in that PR somewhat, with the end-user benefit of displaying time zones in the local time rather than UTC. No accompanying issue.

## Approach

Instead of creating a character vector of format strings based on the format of the time zone, we do an additional regex pass on the incoming timestamps. This lets us use a single format string to parse them.

The new substitution means that UTC time stamps are now explicitly given `+0000` offsets, which means that the `tz` used in `as.POSIXct` can be the user's local time zone. Previously, they were parsed treating the trailing `Z` as a literal, which meant that to R, they did not have a time zone, so all time stamps had to be parsed as UTC. (This is hard to explain!)

The comprehensive tests we added last time we touched this code were modified slightly, and do a great job at validating the new behavior. :)

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,10 @@
 
 ## Enhancements and fixes
 
-- `get_groups()` now paginates through all results when a `prefix` is provided, if the Connect server API version supports pagination. (#328)
+- `get_groups()` now paginates through all results when a `prefix` is provided,
+  if the Connect server API version supports pagination. (#328)
+- Timestamps from the Connect server are now displayed in your local time zone,
+  rather than in UTC. (#400)
 
 # connectapi 0.7.0
 
--- a/R/parse.R
+++ b/R/parse.R
@@ -156,7 +156,7 @@ coerce_datetime <- function(x, to, ...) {
 parse_connect_rfc3339 <- function(x) {
   # Convert timestamps with offsets to a format recognized by `strptime`.
   x <- gsub("([+-]\\d\\d):(\\d\\d)$", "\\1\\2", x)
-  x <- sub("Z$", "+0000", x)
+  x <- gsub("Z$", "+0000", x)
 
   # Parse with an inner call to `strptime()`, which returns a POSIXlt object,
   # and convert that to `POSIXct`.
--- a/R/parse.R
+++ b/R/parse.R
@@ -154,35 +154,21 @@ coerce_datetime <- function(x, to, ...) {
 # - "2020-01-01T00:02:03-01:00"
 # nolint end
 parse_connect_rfc3339 <- function(x) {
-  # Convert any timestamps with offsets to a format recognized by `strptime`.
+  # Convert timestamps with offsets to a format recognized by `strptime`.
   x <- gsub("([+-]\\d\\d):(\\d\\d)$", "\\1\\2", x)
+  x <- sub("Z$", "+0000", x)
 
-  # `purrr::map2_vec()` converts to POSIXct automatically, but we need
-  # `as.POSIXct()` in there to account vectors of length 1, which it seems are
-  # not converted.
-  #
-  # Parse with an inner call to `strptime()`; convert the resulting `POSIXlt`
-  # object to `POSIXct`.
+  # Parse with an inner call to `strptime()`, which returns a POSIXlt object,
+  # and convert that to `POSIXct`.
   #
   # We must specify `tz` in the inner call to correctly compute date math.
-  # Specifying `tz` when parsing just changes the time zone without doing any
-  # date math!
+  # Specifying `tz` when in the outer call just changes the time zone without
+  # doing any date math!
   #
-  # > xlt
-  # [1] "2024-08-29 16:36:33 EDT"
-  # > tzone(xlt)
-  # [1] "America/New_York"
-  # > as.POSIXct(xlt, tz = "UTC")
-  # [1] "2024-08-29 16:36:33 UTC"
-  purrr::map_vec(x, function(.x) {
-    # Times with and without offsets require different formats.
-    format_string <- ifelse(
-      grepl("Z$", .x),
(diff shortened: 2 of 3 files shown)


tests/testthat/test-parse.R
```diff
@@ -109,7 +127,14 @@ test_that("parse_connect_rfc3339() parses timestamps with offsets as expected",
...
 })
 
+
 test_that("parse_connect_rfc3339() handles fractional seconds", {
+  original_tz <- Sys.getenv("TZ")
+  withr::defer({
+    Sys.setenv(TZ = original_tz)
+  })
+
+  Sys.setenv(TZ = "UTC")
```
nealrichardson, line 137: Can we use `local_envvar()` for this?

```suggestion
  withr::local_envvar(TZ = "UTC")
```

tests/testthat/test-parse.R
```diff
@@ -125,7 +150,10 @@ test_that("parse_connect_rfc3339() handles fractional seconds", {
 })
 
 test_that("make_timestamp produces expected output", {
-  withr::defer(Sys.setenv(TZ = Sys.getenv("TZ")))
+  original_tz <- Sys.getenv("TZ")
```
nealrichardson, line 153: Maybe leave a comment above this saying that the test itself will mutate the time zone a few times, and we want to reset?

You could also wrap each block with `withr::with_envvar()`

R/parse.R
```diff
@@ -154,35 +154,21 @@ coerce_datetime <- function(x, to, ...) {
 # - "2020-01-01T00:02:03-01:00"
 # nolint end
 parse_connect_rfc3339 <- function(x) {
-  # Convert any timestamps with offsets to a format recognized by `strptime`.
+  # Convert timestamps with offsets to a format recognized by `strptime`.
   x <- gsub("([+-]\\d\\d):(\\d\\d)$", "\\1\\2", x)
+  x <- gsub("Z$", "+0000", x)
```
nealrichardson, line 159: I don't understand: these two regexes are not going to both be operating on a timestamp because either it ends in `+00:00` or it ends in `Z` but it can't end in both. Is Connect inconsistent in how it emits timestamps such that we would need both?

NEWS.md
```diff
@@ -2,7 +2,10 @@
 
 ## Enhancements and fixes
 
-- `get_groups()` now paginates through all results when a `prefix` is provided, if the Connect server API version supports pagination. (#328)
+- `get_groups()` now paginates through all results when a `prefix` is provided,
+  if the Connect server API version supports pagination. (#328)
+- Timestamps from the Connect server are now displayed in your local time zone,
```
nealrichardson, line 7: Just a gut check, it is desirable to see things in local timestamps and not UTC?

nealrichardson (commented):
Code simplification is nice, just a few questions on there for you

NEWS.md
toph-allen: Yeah, definitely. If I'm in NYC and I say to Connect, "Give me all the records from the Metrics firehose from 10 AM today to 5 PM today", and `connectapi` gives me back a data frame with its time zone column in UTC, it's going to be (1) additional mental effort for me to read the times that are returned and (2) if I wanted to group results by day, it would be grouping by day according to UTC. Local time is a distinct improvement IMHO.

R/parse.R
toph-allen: Connect uses Go's built-in RFC3339 formatter (https://pkg.go.dev/time).

In RFC3339, UTC can be denoted either by `+00:00` or `Z` at the end (https://datatracker.ietf.org/doc/html/rfc3339#section-4.2).

This set of regexes allows us to handle both. (Before this PR we handled by in the `purrr::map_vec()` starting on old line 177.

tests/testthat/test-parse.R
toph-allen: `local_envvar()` everywhere :)

R/parse.R
nealrichardson: Sure, the spec may allow that, but does Connect actually do it differently in different places? 

This change is defensive against it, so it's fine, I would just be concerned if Connect were internally inconsistent, that seems worth solving so any consumer of the API doesn't have to worry with this fussiness.

nealrichardson (approved):
Thanks for doing this. I'm still curious if there's an underlying quirk in Connect that needs fixing, but we don't have to block this PR for that.

toph-allen:
Ok, thanks! My take is that the quirk is in the RFC3339 standard — Connect uses Go's default implementation of that standard for serializing dates.

NEWS.md
jonkeane: I'm not arguing that this is what we should do (and especially not in _this_ PR as it is), but there's another option that folks might believe these times are: _the server's_ timezone. And in your example there, _that_ is frequently most relevant (e.g. for when a report was rendered).

R/parse.R
```diff
@@ -154,35 +154,21 @@ coerce_datetime <- function(x, to, ...) {
 # - "2020-01-01T00:02:03-01:00"
 # nolint end
 parse_connect_rfc3339 <- function(x) {
-  # Convert any timestamps with offsets to a format recognized by `strptime`.
+  # Convert timestamps with offsets to a format recognized by `strptime`.
```
jonkeane, line 157: Not required, but might be nice to have an example of that the timestamps look like coming in and what they look like after these `gsubs`

NEWS.md
toph-allen: Yeah, you're right, that could actually be the most intuitive option. This PR has merged, but — would be open to making that change.

R/parse.R
toph-allen: Good suggestion.

PR #404, opened 2025-05-02
feat: support new metrics firehose api with `get_usage()`

### LINKED ISSUE #390: Support upcoming metrics "firehose" v1 endpoint
Landing in the next Connect release.

toph-allen:
## Intent

Adds support for the metrics "firehose" API.

Fixes #390

## Approach

> Explain how you solved the problem and any tradeoffs.
>
> If applicable:
> - Include guidance for reviewers (rewrites ..., start with ..., commit-by-commit, etc.).
> - Link to notes and discussions about this work.
> - Any relevant information about tests, automated or manual.

## Checklist


.lintr
```diff
@@ -1,7 +1,7 @@
 linters: linters_with_defaults(
   line_length_linter = line_length_linter(120L),
   object_name_linter = object_name_linter(styles = c("snake_case", "symbols", "CamelCase")),
-  cyclocomp_linter = cyclocomp_linter(30L),
+  cyclocomp_linter = NULL, # Issues with R6 classes.
```
jonkeane, line 4: We can also get rid of `any::cyclocomp` here too, yeah? https://github.com/posit-dev/connectapi/blob/e8c807523977158f1817abea1cce02da5097a63e/.github/workflows/lint.yaml#L25-L26

R/connect.R
```diff
@@ -818,6 +818,33 @@ Connect <- R6::R6Class(
       self$GET(path, query = query)
     },
 
+    #' @description Get content usage data.
+    #' @param from Optional `Date` or `POSIXt`; start of the time window. If a
+    #' `Date`, coerced to `YYYY-MM-DDT00:00:00` in the caller's time zone.
+    #' @param to Optional `Date` or `POSIXt`; end of the time window. If a
+    #' `Date`, coerced to `YYYY-MM-DDT23:59:59` in the caller's time zone.
```
jonkeane, line 825: I'm not sure this is doing what we expect with timezones. And actually, I'm not even totally sure what is _intended_. So maybe we should work that out here and then adapt the code to fit? When we send timestamps to Connect with this function do we want them to be transformed to UTC from the caller's local timezone before being sent? Or some other behavior?

One thing to note: If I'm reading this correctly, `make_timestamp()` has slightly different behavior if one sends a non-character than if one sends a character input. Where the character string version will not be parsed and also not be transformed into UTC. And so this function will do the same. 

https://github.com/posit-dev/connectapi/blob/e8c807523977158f1817abea1cce02da5097a63e/R/parse.R#L16-L28

R/parse.R
```diff
@@ -101,6 +105,65 @@ parse_connectapi <- function(data) {
   ))
 }
 
+# nolint start
```
jonkeane, line 108: What linting are we escaping here?

tests/testthat/test-get.R
```diff
@@ -365,3 +374,86 @@ test_that("get_content only requests vanity URLs for Connect 2024.06.0 and up",
     )
   })
 })
+
+test_that("get_usage() returns usage data in the expected shape", {
```
jonkeane, line 378: Could we add a test where we pass in a character that is a formatted timestamp? (if we are intending to support that, of course)

R/parse.R
```diff
@@ -101,6 +105,65 @@ parse_connectapi <- function(data) {
...
+        }
+      },
+      "1",
+      USE.NAMES = FALSE
+    )
+  }
+
+  df2[[col_name]] <- NULL
+  df2
+}
```
jonkeane, line 165: > The data returned from the endpoint includes path and user_agent fields nested under a data field. Without special treatment these are returned a list-column, which is awkward. I initially experimented with `tidyr::unnest()`, but that was slow on the larger datasets returned by this endpoint, so I wrote a custom `fast_unnest_character()` function which runs in about 5% (!) of the time that `tidyr::unnest()` takes.

Thinking about this a bit more: this isn't a _huge_ chunk of code of course, but it is another chunk that we will take on the maintenance of if we go this route. This is another example where having our data interchange within connectapi be all data frames means we have to worry about the performance of json-parsed list responses into data frames and make sure those data frames are in a natural structure for folks to use. If we relied instead on only the parsed list data as our interchange and then gave folks `as.data.frame()` methods, we could defer the (sometimes expense) reshaping until late in process and eaking out performance gains like this are much less important so we can rely on more off the shelf tools.

R/connect.R
```diff
@@ -818,6 +818,33 @@ Connect <- R6::R6Class(
...
+    inst_content_hits = function(from = NULL, to = NULL) {
```
nealrichardson, line 826: I don't think this method name benefits from having the `inst_` prefix on it. 

More of a philosophical musing, but I also am not sure it's worth packing more things into methods on the `Connect` R6 object. `get_usage()` is a pretty thin wrapper around this. We could just put this inside of `get_usage()`, not sure there's much benefit to separating things like this. I know this is a pattern in parts of the package, but that doesn't mean we have to keep doing it.

R/get.R
```diff
@@ -526,6 +526,72 @@ get_usage_static <- function(
...
+#' * `timestamp`: The time of the hit as `POSIXct`.
+#' * `path`: The path of the hit. Not recorded for all content types.
+#' * `user_agent`: If available, the user agent string for the hit. Not
+#'   available for all records.
+#'
+#' @details
+#'
+#' The data returned by `get_usage()` includes all content types. For Shiny
+#' content, the `timestamp` indicates the *start* of the Shiny session.
+#' Additional fields for Shiny and non-Shiny are available respectively from
```
nealrichardson, line 561: Why are the additional fields not included in this endpoint?

R/get.R
```diff
@@ -526,6 +526,72 @@ get_usage_static <- function(
...
+
+#' Retrieve content hits for all available content on the server. Available
+#' content depends on the user whose API key is in use. Administrator accounts
+#' will receive data for all content on the server. Publishers will receive data
+#' for all content they own or collaborate on.
+#'
+#' If no date-times are provided, all usage data will be returned.
+
+#' @param client A `Connect` R6 client object.
+#' @param from Optional `Date` or date-time (`POSIXct` or `POSIXlt`). Only
```
nealrichardson, line 541: If it's optional, what happens if I omit it?

R/get.R
```diff
@@ -526,6 +526,72 @@ get_usage_static <- function(
...
+#' @details
+#'
+#' The data returned by `get_usage()` includes all content types. For Shiny
+#' content, the `timestamp` indicates the *start* of the Shiny session.
+#' Additional fields for Shiny and non-Shiny are available respectively from
+#' `get_usage_shiny()` and `get_usage_static()`.
+#'
+#' When possible, however, we recommend using `get_usage()` over
+#' `get_usage_static()` or `get_usage_shiny()`, as it will be much faster for
+#' large datasets.
```
nealrichardson, line 566: What does "large datasets" mean here?

R/parse.R
nealrichardson: Commented code maybe?

tests/testthat/test-get.R
```diff
@@ -365,3 +374,86 @@ test_that("get_content only requests vanity URLs for Connect 2024.06.0 and up",
...
+        user_guid = c(NA, NA, NA, NA, "fecbd383"),
+        content_guid = c(
+          "475618c9",
+          "475618c9",
+          "475618c9",
+          "475618c9",
+          "b0eaf295"
+        ),
+        timestamp = c(
+          parse_connect_rfc3339("2025-04-30T12:49:16.269904Z"),
```
nealrichardson, line 399: Isn't `parse_connect_rfc3339` vectorized?

tests/testthat/test-get.R
```diff
@@ -365,3 +374,86 @@ test_that("get_content only requests vanity URLs for Connect 2024.06.0 and up",
...
+        paste0(
+          "https://connect.example/__api__/v1/instrumentation/content/hits?",
+          "from=2025-04-01T00%3A00%3A01Z&to=2025-04-02T00%3A00%3A01Z"
+        )
+      )
+
+      # Dates are converted to timestamps with the system's time zone, so for
+      # repeatability we're gonna set it here.
+
+      withr::local_envvar(TZ = "UTC")
```
nealrichardson, line 444: Hmm, see #406 though

.lintr
toph-allen: Done.

R/connect.R
toph-allen: Agreed that I don't think the method name at all benefits from `inst_` prefix; to the reader, it isn't at all clear what it means.

I would also be completely fine with beginning to move everything into the functions; I agree that I don't think we get much benefit at all from having the logic in the R6 object. In fact, I think it _adds_ confusion: it was hard for me to decide where we want some of the cutoff to be.

R/get.R
toph-allen: Why am I having so much trouble coming up with the language for this? 😂 

If it's omitted, the time constraint is unbounded on the lower side.

R/get.R
toph-allen: I mean, not large in the "big data" sense, but like, to fetch 100,000 records on the old API requires 200 requests, which starts to lead to delays of a multiple seconds scaling linearly. Perhaps best to just not describe it?

R/parse.R
toph-allen: Yeah, commented code — it's not `roxygen2` docs, just a regular comment, because it's not an exported function.

R/parse.R
toph-allen: I can see what you mean, and this does align with what you've been saying about other server objects.

I hear what you're saying about parsing to data frames for _data interchange_ and I think that approach would be great to use for, say, the integrations endpoints that I just added stories for.

For the data from the hits endpoint, presenting it as anything other than a data frame goes back to feeling kinda not-R-idiomatic, as it isn't data that can… hmm… 

So definitely _one_ of the tasks, and maybe the main task that _I_ can imagine for this data is to, like, treat it as a data frame and filter, plot, etc., it. But another thing you might want to do is, like, _get the content item associated with this hit_. And yeah, in that case, you might just want to be able to pass the `hit`, or `hit$content_guid` to `content_item()`.

I still think we might want to keep code like this around in an `as.data.frame()` method — for making an unnested data frame out of nested data from the Connect API, `tidyr::unnest()` was _20x_ slower (which I was surprised by! it seems like a wild differential).

Open to a variety of options — let's discuss what the best approach would be to finalize and merge this PR.

R/get.R
toph-allen: Apologies for an extremely belated reply to this comment — I missed it my first time through.

When we were creating the new usage endpoint, we wanted it to be as simple as possible and contain just the unified data that's available for both types of hits.

R/parse.R
toph-allen: Calling `tidyr::unnest_wider(usage, data, ptype = list(path = character(0), user_agent = character(0)))` is much more performant — that would be reasonable. But I don't really want to add a `tidyr` dependency to `connectapi`, I realize — we've been trying to *remove* dependencies. I have conflicting feelings on how to proceed here.

tests/testthat/test-get.R
toph-allen: Yes, good catch.

R/get.R
```diff
@@ -526,6 +526,76 @@ get_usage_static <- function(
...
+
+  usage_raw <- client$GET(
+    v1_url("instrumentation", "content", "hits"),
+    query = list(
+      from = make_timestamp(from),
+      to = make_timestamp(to)
+    )
+  )
+  usage <- parse_connectapi_typed(usage_raw, connectapi_ptypes$usage)
+  fast_unnest_character(usage, "data")
```
nealrichardson, line 597: Perhaps to consider in a followup: is there anywhere else this function should be used? For example, if you `include=owner` in GET `v1/content/`, that adds a list column right?

nealrichardson (approved):
A couple of minor suggestions but otherwise :shipit: 🚀

R/parse.R
karawoo: Maybe an option that splits the difference somewhat would be to have the function return a list, provide an `as.data.frame` method that only works if you have tidyr installed, and make tidyr a suggested dependency but not a required one?

toph-allen: > Maybe an option that splits the difference somewhat would be to have the function return a list, provide an `as.data.frame` method that only works if you have tidyr installed, and make tidyr a suggested dependency but not a required one?

@karawoo I took this approach — take a look and see how it reads to you!

PR #405, opened 2025-05-05
Cache timezone response and remove other caller of unversioned API

nealrichardson:
## Intent

Remove the last (unconditional) caller of the unversioned timezones API.

## Approach

Since timezones won't likely change during the course of an R session (and the only calling of this is for printing progress messages, it seems, so whatever), memoize the timezones data on the connect session object, using the approach of trying the v1 API first. Then remove the two places that were calling the API, one of which was not calling the v1 API.

## Checklist

### DIFF
--- a/tests/testthat/test-parse.R
+++ b/tests/testthat/test-parse.R
@@ -131,7 +131,7 @@ test_that("parse_connect_rfc3339() handles fractional seconds", {
       "2024-12-06T19:09:29.948070345+0000"
     ),
     format = "%Y-%m-%dT%H:%M:%OS%z",
-    tz = "UTC"
+    tz = Sys.timezone()
   ))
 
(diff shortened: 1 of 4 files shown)


tests/testthat/test-parse.R
```diff
@@ -131,7 +131,7 @@ test_that("parse_connect_rfc3339() handles fractional seconds", {
       "2024-12-06T19:09:29.948070345+0000"
     ),
     format = "%Y-%m-%dT%H:%M:%OS%z",
-    tz = "UTC"
+    tz = Sys.timezone()
```
nealrichardson, line 134: https://github.com/posit-dev/connectapi/issues/406

nealrichardson:
Seeing no dissent, I'm going to merge this. If anyone has feedback, happy to follow up.

PR #407, opened 2025-05-05
fix: remove unnecessary null check for oauth cred exchange

mconflitti-pbc:
## Intent

Support ticket was raised: https://rstudioide.zendesk.com/agent/tickets/115510

There was an extra null check added in the sdk that is unnecessary since the connect endpoint handles this check in the latest version. By adding a default value here, version 0.7.0 breaks when trying to do a credential exchange with an older version of connect < 2025.03.0.

This check is not done in posit-sdk-py so this brings them into parity.

## Approach

Removes the null check and default for the `requested_token_type`. So if the value is not set, a null value is passed through the api which will then determine which the proper default is in both older and newer versions of connect.

## Checklist


mconflitti-pbc:
@toph-allen Updated the tests like we did last time since the hashes changed for those two functions. Is this adequate? I am not familiar enough with how to build this and use it locally to verify e2e.

tdstein (approved):

jonkeane (approved):
Thanks for this!

mconflitti-pbc:
Verified this fixes the cred exchange call on older version of the connect API.

PR #409, opened 2025-05-08
refactor: use v1 keys API in integration test setup

nealrichardson:
## Intent

Removes the last caller of this unversioned API. Only used in the integration tests.

## Approach

Switch to `v1/users/guid/keys` where available, falling back to the old one where it doesn't. Some other tidying while in there.

If the integration tests pass, this proves that the fallback works. But until we do #395, we don't have a test run that exercises this. For better or worse, this is test only code.

### DIFF
--- a/R/utils-ci.R
+++ b/R/utils-ci.R
@@ -221,9 +221,15 @@ create_first_admin <- function(url, user, password, email) {
 
   client$login(user = user, password = password)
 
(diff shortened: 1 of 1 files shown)


karawoo (approved):

toph-allen (approved):
Thanks for tackling this — I think it looks good to go.

PR #412, opened 2025-05-09
Respect `include` argument when querying for content with GUID

### LINKED ISSUE #411: `get_content()` discards tags when guid parameter is provided
This is kind of surprising?

``` r
library(connectapi)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

client <- connect()
#> Defining Connect with server:

CONTENT_GUID <- "8cd033a0-7e0c-415d-8136-9da1a755c63c"

target_content <- content_item(client, CONTENT_GUID)
get_content_tags(target_content)
#> Posit Connect Tag Tree (content)
#> ├── FK_cat2
#> │   └── sub2
#> └── Action
#>    └── Fast and Furious

with_guid <- get_content(client, guid = CONTENT_GUID)
"tags" %in% names(with_guid)
#> [1] FALSE

no_guid <- get_content(client) |> filter(guid == CONTENT_GUID)
"tags" %in% names(no_guid)
#> [1] TRUE
```

<sup>Created on 2025-05-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

karawoo:
## Intent

`Connect$content()` has an argument `include` where additional query parameters can be specified, but when `guid` was provided we were not respecting the `include` arg.

Fixes #411

## Approach

Included the `include` value when querying by GUID.

## Checklist


karawoo:
ok, this has more downstream impacts than I realized so marking as a draft for now while I sort it out.

tests/testthat/test-content.R
```diff
@@ -168,6 +168,12 @@ without_internet({
       con$content(name = "A name for content"),
       "https://connect.example/__api__/v1/content?name=A%20name%20for%20content&include=tags%2Cowner"
     )
+
+    uuid <- uuid::UUIDgenerate()
```
nealrichardson, line 172: This is fine, but FTR I don't think you need an actual guid here, you could pass `guid = 1` and assert that you `GET https://connect.example/__api__/v1/content/1?include=tags,owner`

R/content.R
```diff
@@ -32,7 +32,7 @@ Content <- R6::R6Class(
     },
     #' @description Obtain the content data from the Connect server.
     get_content_remote = function() {
-      new_content_details <- self$get_connect()$content(self$get_content()$guid)
+      new_content_details <- self$get_connect()$content(self$get_content()$guid, include = NULL)
```
nealrichardson, line 35: Why `include = NULL`? I know we weren't including it before, but it's probably not bad to have it. Might let us do some other improvements like https://github.com/posit-dev/connectapi/issues/402.

Not saying you have to change this, but if you leave it as is, it might be nice to leave a comment/todo explaining that we don't _need_ to have this as `NULL`, just so we don't assume that it must be this way in the future.

tests/testthat/test-content.R
```diff
@@ -168,6 +168,12 @@ without_internet({
...
+    expect_GET(
+      con$content(guid = uuid, include = "tags"),
+      glue::glue("https://connect.example/__api__/v1/content/{uuid}?include=tags")
```
nealrichardson, line 175:
```suggestion
      glue::glue("https://connect.example/__api__/v1/content/{uuid}?include=tags,owner")
```

nealrichardson (approved):
Thanks for jumping on this!

tests/testthat/test-content.R
karawoo: I'm explicitly passing just `"tags"` to `include` here so I think it's correct as written, unless you're suggesting I also change line 174?

nealrichardson: Oh sorry, you're right, I was thinking about the default value (maybe that's worth a test too), and how the matching in `expect_GET` doesn't require you to specify the exact output, it does substring matching.

R/content.R
karawoo: It was mainly to avoid unintentionally changing existing behavior (and also because changing it resulted in a bunch of test failures that I didn't fully understand, but now do thanks to @jonkeane's help). I've gone ahead and removed the `include = NULL` both here and at line 593, and updated the test cases accordingly. The unit tests now pass on my machine but seem to be failing on CI so I will need to investigate a little further.

karawoo: oh...one of the test files didn't get picked up by git as having been renamed so the new version was just hanging out as an untracked file on my machine. hopefully this should be fixed now.

tests/testthat/test-content.R
karawoo: oh good to know. I'll add a test for the default value and update this one to explicitly pass an argument that doesn't match the default (maybe just swapping the order)

tests/testthat/test-content.R
```diff
@@ -168,6 +168,16 @@ without_internet({
       con$content(name = "A name for content"),
       "https://connect.example/__api__/v1/content?name=A%20name%20for%20content&include=tags%2Cowner"
     )
+
+    expect_GET(
+      con$content(guid = 1),
+      glue::glue("https://connect.example/__api__/v1/content/1?include=tags%2Cowner")
```
nealrichardson, line 174:
```suggestion
      "https://connect.example/__api__/v1/content/1?include=tags%2Cowner"
```

tests/testthat/test-content.R
```diff
@@ -168,6 +168,16 @@ without_internet({
...
+    )
+
+    expect_GET(
+      con$content(guid = 1, include = "owner,tags"),
+      glue::glue("https://connect.example/__api__/v1/content/1?include=owner%2Ctags")
```
nealrichardson, line 179: We don't need to glue anymore, right?

```suggestion
      "https://connect.example/__api__/v1/content/1?include=owner%2Ctags"
```

nealrichardson (approved):
LGTM

tests/testthat/test-content.R
karawoo: whoops thanks for the catch

PR #415, opened 2025-05-16
refactor: remove get_apps() and usage of undocumented applications API

nealrichardson:
## Intent

Remove the last caller of the undocumented `GET /__api__/applications/` from this library. 

## Approach

`get_apps()` was the only caller of this API, and it was only being used in two places, and only to check whether a content item with a given name already exists. The `GET /__api__/v1/content/` API accepts a `name` query parameter, so we could swap those to use that.

I also deleted the `dashboard_url_chr()` function, which oddly was exported, because content items include `dashboard_url` in them. 

## Checklist


toph-allen (approved):
Thank you @nealrichardson! Love to see code getting deleted, especially when it calls undocumented APIs.

It feels to me like `get_apps()` was a relic of an earlier time in `connectapi`, before `get_content()`. Perhaps left in as a way to maintain backward compatibility? Makes me wonder about other old or redundant code that's lying around.

Either way, looks good to me, and I see that you got CI working! Thank you!

nealrichardson:
> Either way, looks good to me, and I see that you got CI working! Thank you!

Yes, it looks like in older versions of Connect, if you specified both `name` and `include` as query params, the v1 content listing API would 500. We don't need the includes though for this request so it's all good.

toph-allen:
> Yes, it looks like in older versions of Connect, if you specified both `name` and `include` as query params, the v1 content listing API would 500. We don't need the includes though for this request so it's all good.

Huh, that's… good to know! Particularly as I had been thinking about turning on `includes` for the `content_item()` function (because it's on by default for `get_content()`).

nealrichardson:
It might be fine! Only thing to do is try it :D

PR #417, opened 2025-05-20
refactor: clean up integration tests

### LINKED ISSUE #308: chore: update integration tests to 2024.09.0

### LINKED ISSUE #395: chore: update integration testing matrix

nealrichardson:
## Intent

To get integration tests running on newer versions of Connect, we will need to adapt the example content that uses R to be able to work with different versions of R. Currently this is not working because the manifests are checked in, so the version of R and the package dependencies is fixed. One idea discussed on #396 is to use `rsconnect` to generate the manifests in the test runs themselves, and we can be sure to run the tests for each version of Connect using a version of R that is present in the Docker image so that we can ensure they match. 

This PR doesn't go all the way to that, but it does clean up and refactor the code that creates test content. And it reorganizes the tests so that it is easier to conditionally skip the tests that use a shiny or rmd example (there aren't many, as it turns out), so that we can get most of the CI running on newer Connect versions in the meantime.

Fixes #308
Fixes #395

## Approach

* Audit which example content was being used where. Move some nonfunctioning examples that are used for input validation in `bundle_*` functions to a different directory.
* Delete a couple of content items that were either not used, or were uploaded in a test file but never asserted against.
* Add a `deploy_example()` helper function that wraps the places we reference the remaining examples with code. This made it a little easier to see where they were being called, and will be the place in which we can insert the `rsconnect` business in a future step.
* Break out the test blocks that use the shiny app or the rmd from the big `test-content.R` to their own files, and add a `skip_if` at the top of the file so that they don't run on the newer versions of Connect.
* Cherry-pick the commit that adds the new versions to the CI matrix.
* Clean up, fix stuff that broke, etc. Now we have mostly running integration tests on the latest versions of Connect 🎉


toph-allen:
Following this with interest — thanks for digging into this @nealrichardson!

tests/integrated/helper.R
```diff
@@ -13,3 +13,28 @@ skip_if_connect_older_than <- function(client, version) {
...
+    connect = connect,
+    bundle = bund,
+    name = name,
+    ...
+  )
+
+  guid <- tsk$get_content()$guid
+  content <- content_item(tsk$get_connect(), guid)
+
+  # TODO: a smarter, noninteractive wait...
```
nealrichardson, line 37: Not my todo, just carrying it over from the source where I copied this from in the tests.

tests/integrated/test-deployment.R
```diff
@@ -8,7 +8,8 @@ test_that("can create content", {
   expect_equal(cont1$name, cont1_name)
   expect_equal(cont1$title, cont1_title)
 
-  get_cont1 <- test_conn_1$content(guid = cont1$guid)
+  # Use include = NULL because we aren't passing the include in content_create()
+  get_cont1 <- test_conn_1$content(guid = cont1$guid, include = NULL)
```
nealrichardson, line 12: This test failed before I added this, but I could not reproduce it locally for some reason. That said, I think it's fine to do because `content_create()` doesn't result in a fully hydrated content object anyway, it only creates the entity record, and some fields like about the runtime aren't populated until you deploy a bundle. So the content record you get back from that `POST` won't ever be all that useful (except as a source of the guid of the record that was created).

.github/workflows/integration-tests.yaml
```diff
@@ -15,6 +15,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - "2025.03.0" # jammy
+          - "2024.09.0" # jammy
```
toph-allen, line 19: So good to see! 🎉

tests/integrated/test-env-vars.R
```diff
@@ -0,0 +1,132 @@
+# Temporarily skip these on newer Connect versions because the manifest for this
+# content specifies an old version of R not found in newer images.
+skip_if(safe_server_version(test_conn_1) > "2024.03.0")
+
+# ensure that RSPM is being used so these do not take eternity
+rmd_content <- deploy_example(test_conn_1, "rmarkdown")
```
toph-allen, line 6: If we wanted to maintain these tests across the full set of tested versions, we could just duplicate the `rmarkdown` content with different manifests, and condition the content name on `safe_server_version(test_conn_1)`.

toph-allen (approved):
This looks good! Thanks again for tackling it. Could you create a follow-up issue to generate new manifests for test content items with newer R versions?

tests/integrated/test-env-vars.R
nealrichardson: I'll make a note of that as a possibility in the followup issue, an alternative to using rsconnect.

nealrichardson:
Followup issue: https://github.com/posit-dev/connectapi/issues/418

PR #423, opened 2025-07-15
feat: added support for optional audience param

mconflitti-pbc:
## Intent

Adds optional audience parameter to our oauth creds helpers. Unclear how to handle connect versions constraint since that relies on the oauth feature being there but this is an added param not a new function.

Need to address parity the python sdk through additional integration and association helpers as a follow up.

## Approach

Add optional param

## Checklist


R/get.R
```diff
@@ -671,15 +674,17 @@ get_procs <- function(src) {
...
 ) {
   validate_R6_class(connect, "Connect")
   url <- v1_url("oauth", "integrations", "credentials")
   body <- list(
     grant_type = "urn:ietf:params:oauth:grant-type:token-exchange",
     subject_token_type = "urn:posit:connect:user-session-token",
     subject_token = user_session_token,
-    requested_token_type = requested_token_type
+    requested_token_type = requested_token_type,
+    audience = audience,
```
nealrichardson, line 687:
```suggestion
    audience = audience
```

R/get.R
```diff
@@ -745,7 +754,8 @@ get_oauth_content_credentials <- function(
     grant_type = "urn:ietf:params:oauth:grant-type:token-exchange",
     subject_token_type = "urn:posit:connect:content-session-token",
     subject_token = content_session_token,
-    requested_token_type = requested_token_type
+    requested_token_type = requested_token_type,
+    audience = audience,
```
nealrichardson, line 758:
```suggestion
    audience = audience
```

R/get.R
```diff
@@ -806,12 +819,13 @@ get_oauth_content_credentials <- function(
...
 #' @export
-get_aws_credentials <- function(connect, user_session_token) {
+get_aws_credentials <- function(connect, user_session_token, audience = NULL) {
   error_if_less_than(connect$version, "2025.03.0")
   response <- get_oauth_credentials(
     connect,
     user_session_token,
-    requested_token_type = "urn:ietf:params:aws:token-type:credentials"
+    requested_token_type = "urn:ietf:params:aws:token-type:credentials",
+    audience = audience,
```
nealrichardson, line 828:
```suggestion
    audience = audience
```

R/connect.R
```diff
@@ -1017,7 +1021,8 @@ connect <- function(
       visitor_creds <- get_oauth_credentials(
         con,
         user_session_token = token,
-        requested_token_type = "urn:posit:connect:api-key"
+        requested_token_type = "urn:posit:connect:api-key",
+        audience = audience,
```
nealrichardson, line 1025:
```suggestion
        audience = audience
```

R/get.R
```diff
@@ -745,7 +754,8 @@ get_oauth_content_credentials <- function(
     grant_type = "urn:ietf:params:oauth:grant-type:token-exchange",
     subject_token_type = "urn:posit:connect:content-session-token",
     subject_token = content_session_token,
-    requested_token_type = requested_token_type
+    requested_token_type = requested_token_type,
+    audience = audience,
   )
   connect$POST(
```
nealrichardson, line 760: Two questions: 

* Should we check that the connect server version is new enough if `!is.null(audience)`? I'm guessing that if you want to specify audience and you're on Connect older than the upcoming release, you won't get back what you're asking for.
* Does the `POST` handler on the server ignore unknown body fields? Like, if you were to include `audience` in the payload for an older Connect.

Assuming that the answers are "yes" and "no", respectively, I think we handle this like this here, instead of adding it in the `body` definition above:

```r
if (!is.null(audience)) {
  if (connect$version < something) {
    stop(helpful message that this is not supported)
  }
  body$audience <- audience
}
```

Note that the tests in `tests/testthat/` don't test against real Connect servers, so they won't give us any assurance about how older versions of Connect respond if you include this parameter. There is an integration test suite that runs against a range of Connect versions but IDK how feasible it would be to set them up with integrations to test this.

nealrichardson (changes requested):
We can chat about testing when you're back if you like. I think the approach will be to just assert that we're making the expected POST requests.

toph-allen (changes requested):
Feedback in two major categories:

- Thoughts on the workflow generating a publisher-scoped client object, which depend upon how the `audience` parameter interacts with `requested-token-type` (which I'm not 100% sure about).
- Comments on documentation / how we communicate the parameter.

---

I realize now in your Slack comment about the testing framework — the easiest way to bootstrap those responses you need is to capture them from a Connect server. Happy to pair on that if you want. When you capture them from a server, appropriately-named `.json` files are created.

R/get.R
mconflitti-pbc: Do we do this version checking elsewhere?

tests/testthat/test-oauth.R
```diff
@@ -148,4 +148,222 @@ with_mock_api({
...
+    )
+    expect_equal(
+      credentials,
+      list(
+        access_token = "content-access-token",
+        issued_token_type = "urn:ietf:params:oauth:token-type:access_token",
+        token_type = "Bearer"
+      )
+    )
+  })
```
toph-allen, line 272: I don't know that this test and e.g. the test on line 207 really exercise different parts of the code (same for the other tests that follow this format). Because `MockConnect` doesn't enforce expectations about the contents of the request body, we're basically confirming that the correct route is called and the expected data is returned.

I personally think it'd be worth adding that capability in the future, as for tests like this, I find it easier to have the expectations defined alongside the tests than in JSON files that are loaded implicitly from the requests.

Perhaps it makes sense to simplify these tests for now, removing a few of them.

It might also make sense to create an issue to track adding the ability to provide an expected request body in `MockConnect$mock_response`.

toph-allen (commented):
This generally looks good to me, just a few comments about tidying up tests etc.

I'm holding off on hitting "Approve" because I think it'd be good to validate this on a real Connect server. I'm not sure how we could do that easily on an integration test; it might require coding a bit of toy content and deploying it to a dev Connect instance to ensure that the audience parameter works.

Perhaps I'm being overly cautious, and if so, @mconflitti-pbc or other reviewers can push back. I'm happy to help verify this tomorrow. :)

toph-allen (commented):
Oops, I meant my last review to go through as "Comment", not "Request changes".

[edit] I guess I _did_ select comment. At least, I know I did for _this_ review. Does GH automatically show your review as "Requested changes" if you leave a comment with suggestions and select "Comment" now?

toph-allen (approved):
Approving after verifying on the server.

tests/testthat/2024.08.0/__api__/v1/oauth/integrations/credentials-224fa9-POST.json
```diff
@@ -0,0 +1,5 @@
+{
+    "access_token": "visitor-api-key",
+    "issued_token_type": "urn:posit:connect:api-key",
+    "token_type": "Key"
+}
```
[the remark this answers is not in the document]
mconflitti-pbc: oh thats right! my apologies you did mention that already

PR #441, opened 2025-08-11
feat: get content that uses a specified integration

### LINKED ISSUE #433: Add function to get content associated with an integration
Recipe: [Finding Content Associated with an Integration](https://docs.posit.co/connect/cookbook/oauth-integrations/finding-associations-by-integration/index.html)

Add a function that takes an integration and returns a list of content items it's associated with.

It would be nice to be able to write:

```r
client <- connect()
integrations <- get_integrations(client)
github_integration <- integrations[1]
get_content(github_integration)
```

To accomplish that, we'd have to do two things:
1. Add the Connect client object. Nice to do it as an attribute maybe. Easy.
2. Uh… a function named `get_content()` already exists, with a very specific signature that kinda precludes doing something like `get_content(x)`. The first argument is named `src`.

Ways forward:

- Switch to method dispatch for `get_content()` but keep the weird name of the first argument `src`. People will mostly write `get_content(integration)` anyway.
- Eat the pain of changing the name of the first argument to `get_content()` now. Most people don't pass in the client with a name, I would guess. Use method dispatch.
- Name this function something like `get_content_by_integration()`. At a later date, make the painful breaking change in `get_content()`. (Maybe at a point when we're making more breaking changes?)

toph-allen:
## Intent

Get a list of content that is associated with a specific integration.

Fixes #433

## Approach

- Created a new function named `get_content_list()`. It takes a single parameter, an integration, and returns a list of content objects.
  - For function naming and return type, I discussed the tradeoffs of various approaches with @jonkeane and this was the least-worst option we arrived at (see comment [here](https://github.com/posit-dev/connectapi/issues/433#issuecomment-3176838548)).
  - I named the argument `x`, leaving it open to accepting e.g. a Connect client object in a future PR.
- Mostly was able to use existing fixtures for the tests.

## Checklist


R/integrations.R
```diff
@@ -286,6 +286,57 @@ set_integrations <- function(content, integrations) {
...
+  }
+  client <- attr(x, "client")
+  error_if_less_than(client$version, "2024.12.0")
+  assoc <- client$GET(v1_url(
+    "oauth",
+    "integrations",
+    x$guid,
+    "associations"
+  ))
+  purrr::map(
```
nealrichardson, line 334: This could be a potentially large list, so I'm not sure how wise this.

I wonder if a better way to expose this would be through the content search, adding an `integrations:guid` keyword. There would be other benefits there too, but if nothing else, that would give you a list of content items already, so you wouldn't have to iterate over guids.

Alternatively, we could expose a different endpoint off of `oauth/integrations/guid/` that would return a list of content records. But IMO running this through search would be better than adding a new endpoint.

NEWS.md
```diff
@@ -9,6 +9,8 @@
 - New functions allow you to manage the OAuth integrations on your Connect
   server: `create_integration()`, `update_integration()` and
   `delete_integration()`. (#434)
+- New `get_content_list()` function retrieves a list of content items associated with
```
nealrichardson, line 12: Doing something specific to an oauth integration is not what I would expect a function called `get_content_list()` to do.

mconflitti-pbc: yeah i would agree here. i think naming could be an issue. I almost always prefer explicit naming conventions like: `get_associated_content_for_integration()` or something like that. `get_content_list` is too similar to what I would expected to get a list of content objects.

toph-allen:
Coming back to this after PTO — I'm either going to rename the functions slightly and keep the somewhat janky "call the content endpoint multiple times", or punt on this until implementing it under the search endpoint and implementing a `search_content()` function in `connectapi`. Probably the first one now, and the second one in the future.

toph-allen:
After discussing with @jonkeane this morning, I'm closing this pull request. This isn't a feature that people are *clamoring* for, and it would be better served by the search endpoint. I'm going to create follow-up issues in both `connect` and `connectapi` to track.

Issue #272, opened 2024-05-24
Support content search query


nealrichardson:
Need to check for Connect version and only use it if > 2024.03.0, else maybe warn and fall back to v1/content.

Can easily remove `get_apps()` after this, the remaining calls to it after #271 are just to find a content item by name.

PR #444, opened 2025-09-23
feat: simple function to wrap `GET v1/search/content`

toph-allen:
## Intent

Implement a first pass at content search.

Partially addresses #272

## Approach

- A basic function to call the content search API, just passing in the query string directly.
- Just returns the raw server response currently.
- Does not wrap pagination; provides the pagination parameters as function arguments. This may be addressed in a future PR.

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,7 @@
 - New functions allow you to manage the OAuth integrations on your Connect
   server: `create_integration()`, `update_integration()` and
   `delete_integration()`. (#434)
+- Support content search API with the `search_content()` function. (#272)
 
 # connectapi 0.8.0
 


R/content.R
```diff
@@ -1434,3 +1434,28 @@ get_content_packages <- function(content) {
   res <- content$packages()
   parse_connectapi_typed(res, connectapi_ptypes$content_packages)
 }
+
+search_content <- function(client, terms = character()) {
```
nealrichardson, line 1438: If this is the lower-level API wrapper (which I agree has some value), I'd make the function args match the query parameters, and for good measure I'd include `...` for forward compatibility. Like: 

```suggestion
search_content <- function(client, q = NULL, include = "owner,vanity_url", limit = Inf, page_size = 500, page_number = 1, ...) {
```

That also makes the rest of the function dead simple.

Also IMO, I would just export this now, and in fact would declare victory and move on without doing the "Add structured arguments for filters" task you outlined--if you can search using the same search syntax you can do in the web UI, that's really powerful, and I'm not sure we would need the structured arguments for filters. I'd wait until we got feedback or saw for ourselves that we really needed more arguments, and then if it turns out that we do, we can add them as arguments to this function (or handle them in the `...`).

toph-allen: I'm sold on half of what you say and unsure about the other half (gonna think through it here).

The part I'm sold on is passing in just the query string. We have a good query syntax that we put a lot of work into, and building additional machinery on the R side is making extra work for myself.

I'm not sold on just having the page parameters be arguments at the top level here — I would ideally like to have this function to abstract away pagination and have the user facing function just return all of the results. I could see an argument for simplicity's sake, but perhaps that isn't what you were thinking.

nealrichardson: I agree we should automatically page, but if for some reason you want to change the default behavior, including them in the function signature gives you a way out.

toph-allen: For now, since this is a simple wrapper, I'm omitting paging (even though I think it is probably a requirement for this to release). Leveraging the built-in paging mechanism would tie us very closely to `connectapi`'s existing implementation (and require more iteration cycles on this PR). Going to try this out in the dev version in some existing use cases to learn about how the abstraction could be improved.

Filed https://github.com/posit-dev/connect/issues/33764 to track allowing a larger `page_size`.

PR #449, opened 2025-10-02
feat: add pagination to `search_content()`

### LINKED ISSUE #447: feat: add auto-pagination to `search_content()` function
The `search_content()` function currently exposes the `page_number` and `page_size` arguments, the latter of which maxes out at 500. Functionally, this means that it's not usable without wrapping it in a function that fetches all pages.

Considering two approaches.

**Option 1: Use the current `page_offset` function for pagination.** `page_offset` passes the `page_number` query param as a function argument, which demands that the endpoint is wrapped in a function that implements it (like the current `search_content` function).

With the current `search_content` function, this looks like so:

```r
search_all_content <- function(client, q = NULL, include = "owner,vanity_url") {
  page_offset(client, req = search_content(client, q = q, page_number = 1, page_size = 500, include = include))
}
```

I don't actually propose the outer function be named `search_all_content` — it should be named `search_content`, and the inner wrapper should have a different name. Previously, those inner wrapping funcs have been methods on the `Connect` R6 class. We've been moving away from that. Maybe we want to name the inner function directly for the API path, e.g. `v1_search_content()`.

**Option 2: Use a different approach to paginating.**

If we don't want to _require_ an inner function wrapping the endpoint, we would need a different approach for pagination logic. If the pagination function didn't increment the `page_number` argument to an R call, but instead incremented that parameter in a list, we could iterate directly on the `Connect$GET()` call to the endpoint. This would be a little more work.

toph-allen:
## Intent

Add auto-pagination to `search_content()`.

Fixes #447

## Approach

- Moved the raw API call to an inner function, which is called with `page_offset()`.

## Checklist


R/content.R
```diff
@@ -1587,21 +1581,38 @@ get_content_packages <- function(content) {
 search_content <- function(
   client,
   q = NULL,
-  page_number = 1,
-  page_size = 500,
```
toph-allen, line 1591: I figured we don't need to expose these to users, since pagination is kind of an implementation detail here.

jonkeane: > But retaining them and passing them down might make it easier to test performance under different values

Yeah, that's true. A few thoughts: (1) one _should be able to_ pass these down in `...` even if they are not included in the signature explicitly. (2) what do our other functions do? If they expose these explicitly so that it's obvious there are nobs to tune here, I'm fine including them. But also fine to not (so long as someone can pass them in `...`)

R/content.R
```diff
@@ -1587,21 +1584,40 @@ get_content_packages <- function(content) {
...
   include = "owner,vanity_url",
   ...
 ) {
   error_if_less_than(client$version, "2024.04.0")
 
-  path <- v1_url("search", "content")
+  inner_search <- function(
```
jonkeane, line 1592: I know I called it this in my quick snippet, but what about this: 

```suggestion
  .search_content <- function(
```

toph-allen: Sure!

tests/testthat/2025.09.0/__api__/v1/search/content-d37a6d.json
```diff
@@ -0,0 +1,118 @@
...
+            "id": "57137",
+            "owner": {
+                "guid": "c2250bb4",
+                "username": "toph_allen",
+                "first_name": "Toph",
+                "last_name": "Allen"
+            },
+            "public_content_status": null
+        }
+    ]
```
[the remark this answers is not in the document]
toph-allen: I think that's totally fine! I feel like for this test we could remove even more — they could probably _technically_ be empty except for the fields that are specifically checked in the test…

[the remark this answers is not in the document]
toph-allen: Yeah, but tbh, the hashing also makes it _really_ difficult to know what mock a given test will uses (unless the test is failing to find the mock) — particularly after the fact, and especially when there are other tests that call the same endpoint with different parameters. It's a big usability issue imho.

toph-allen:
> I disagree on passing the parameters with `...`. My opinion is that parameters that go through `...` should get passed to exported functions with their own documentation. If the only place they're documented is through the `...` documentation for this function, to me that just obfuscates them unnecessarily. If we're going to support people providing pagination-related arguments then let's give them informative names that can pop up in tab-completion, etc. 

@karawoo Good point. It does have a bad feel to it.

The other functions that wrap paginated endpoints don't provide pagination controls, and my *original* intent for this func was to do the same. Added them back in because we may want to use them to test performance of different `page_size`s.

Your point makes me feel that including them in `...` is probably a bad compromise between my original intent and current functionality. But the behavior of `page_number` is funky and I have #450 to fix that — I'd like to postpone making the params explicit until that issue.

PR #462, opened 2025-10-14
feat: `connect_user` S3 class

karawoo:
Creates new S3 classes `connect_user` and `connect_users`. Functions that previously took a user GUID as an argument now can take either the GUID or the `connect_user` object. 

This is still a draft while we talk through the right approach for phasing in this new object and moving away from the data frame-centric return object of `get_users()`.

- For now `get_users()` still returns a data frame.
- `get_users_list()` returns a list of `connect_user` objects; `get_users()` just calls `get_users_list()` and converts to a data frame. I don't love this naming at all.
- Arguments that were previously named `user_guid`, `owner_guid`, etc. are now `user`/`owner` to reflect the fact that they can be either the GUID or object. This may be a pretty significant breaking change but it also doesn't seem right to keep `_guid` in there when it might not be a guid. 

I'm a little tempted to deprecate passing the GUID strings with a warning, and then a few releases down the line remove support for passing the GUID. At the same time we can maybe swap `get_users()` to returning a list and remove/deprecate `get_users_list()`. Would love others' thoughts as they look at this though, I think there are a lot of potential options.

## Checklist


R/user.R
```diff
@@ -32,3 +32,28 @@ user_guid_from_username <- function(client, username) {
...
+#'
+#' @param user Either a character string containing a user GUID or a
+#'   `connect_user` object (as returned by `Connect$user()` or `Connect$users()`)
+#'
+#' @return A character string containing the user GUID
+#'
+#' @keywords internal
+get_user_guid <- function(user) {
+  if (is.character(user)) {
+    return(user)
```
jonkeane: Do we want to add a check here that the character is GUID-like?

Issue #461, opened 2025-10-14
`get_content()` documentation and prototype need to be updated with new fields


toph-allen:
The fields described in the documentation for `get_content()` do not contain all the current fields on the `content` object type (see [here](https://docs.posit.co/connect/api/#get-/v1/content/-guid-)).

The list of these fields might be better documented in the `Content` class (either way, we should probably only have the list in one place in `connectapi` for internal consistency's sake.

PR #463, opened 2025-10-14
feat: return a list of `Content` class objects from `search_content()`

### LINKED ISSUE #448: feat: Add better printing to the output of `search_content()`
It's unwieldy to see the entire output of the content search printed as a list of lists. Modifying the list so that every item has a class with a nicer print method would improve this.

### LINKED ISSUE #452: `search_content()` function could return a list of `Content` class objects
The `search_content()` function currently returns a list of lists, with each inner list representing a content item.

Converting those inner lists to the existing R6 `Content` class would allow us to leverage all of the existing work done on that class.

toph-allen:
## Intent

Instead of returning a list of lists from `search_content()`, return a list of `Content` class objects. This allows users to use existing content-related functions with the list via `purrr`.

Fixes #452
Fixes #448 

## Approach

This is pretty minimal so far — just changing the class of the returned object.

Using `purrr` to act upon the list of content is not an ideal workflow, but this is the first step towards something more user-friendly. See the original issue.

Documentation is not ideal. The `Connect` class does not document available fields, and the fields documented on `get_content()` are out of date. I filed #461 to track this.

## Checklist

- [ ] Does this change update `NEWS.md` (referencing the connected issue if necessary)? **[Not needed]**

### DIFF
--- a/R/content.R
+++ b/R/content.R
@@ -1522,7 +1522,13 @@ get_content_packages <- function(content) {
 #' library(connectapi)
 #' client <- connect()
 #'
-#' search_content(client, q = "")
+#' my_content <- search_content(client, q = "owner:@me")
+#'
+#' shiny_content <- purrr::keep(my_content, function(x) {
+#'   x$content$app_mode == "rmd-shiny"
+#' })
+#'
+#' purrr::map(shiny_content, lock_content)
 #' }
 #'
 #' @family content functions
--- a/man/search_content.Rd
+++ b/man/search_content.Rd
@@ -47,7 +47,13 @@ information.
 library(connectapi)
 client <- connect()
 
-search_content(client, q = "")
+my_content <- search_content(client, q = "owner:@me")
+
+shiny_content <- purrr::keep(my_content, function(x) {
+  x$content$app_mode == "rmd-shiny"
+})
+
+purrr::map(shiny_content, lock_content)
 }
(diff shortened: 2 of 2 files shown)


R/content.R
```diff
@@ -1637,7 +1522,13 @@ get_content_packages <- function(content) {
 #' library(connectapi)
 #' client <- connect()
 #'
-#' search_content(client, q = "")
+#' my_content <- search_content(client, q = "owner:@me")
+#'
+#' shiny_content <- purrr::keep(my_content, function(x) {
+#'   x$content$app_mode == "rmd-shiny"
+#' })
```
jonkeane: This is a bit of an odd example, no? I suspect it's there to show that (currently) one should use `purrr::keep(...)` for filtering, but in this case, you would _actually_ want to do `my_content <- search_content(client, q = "owner:@me type:rmd-shiny")` to do this. 

There's an arc of work to make the list more-easily-filterable that we should take up (later/next), but IMO we should document the stuff we're building now with how it should be used today, even if there might be other ways to do / other tasks folks might do with stuff like this in the future.

toph-allen: Good point!! Perhaps I should use the longer example I used in GitHub PRs that uses a comparison that isn't possible in the filter API at the moment.

[edit] And/or just remove this example.

PR #466, opened 2025-10-15
refactor: remove unnecessary get_connect(), get_content(), get_variant() R6 methods

nealrichardson:
## Intent

Just access the R6 slots directly. Remove ambiguity about which pattern we want to use (https://github.com/posit-dev/connectapi/pull/459#discussion_r2430053479) 

## Approach

Find and replace, with some careful reading to distinguish some confusing cases:

* Where `get_connect()` was called on a Connect object itself (it just returned `self`!) rather than some API entity
* Disambiguating the `get_content()` top-level function that retrieves all content records, vs. the `$get_content()` R6 method.

## Checklist

### DIFF
--- a/R/schedule.R
+++ b/R/schedule.R
@@ -50,8 +50,8 @@ VariantSchedule <- R6::R6Class(
       if (self$is_empty()) {
         params <- purrr::list_modify(
           params,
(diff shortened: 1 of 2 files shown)


toph-allen (approved):
:shipit:

PR #468, opened 2025-10-15
fix: update `get_content` ptype and docs

### LINKED ISSUE #461: `get_content()` documentation and prototype need to be updated with new fields

toph-allen:
## Intent

Update the `ptype` and documentation for `get_content()` to include recently-added columns. The `ptype` update means that the columns will be parsed with proper types.

Fixes #461 

## Checklist

### DIFF
--- a/man/bundle_dir.Rd
+++ b/man/bundle_dir.Rd
@@ -21,7 +21,7 @@ Bundle A bundle object
 Creates a bundle from a target directory.
 }
 \examples{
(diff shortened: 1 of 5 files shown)


karawoo (approved):
since we have a separate issue for removing this together, it seems good to keep this up to date as long as we are maintaining it

PR #470, opened 2025-10-16
feat: add functions to convert content lists to data frames

### LINKED ISSUE #451: Add way to convert list of content objects from `content_search()` into a data frame
`search_content()` returns a list of content objects from the server (currently as a plain list). It's useful to turn this into a data frame sometimes.

Could add `as.data.frame()` and `as_tibble()` methods like what we did for integrations.

https://github.com/posit-dev/connectapi/blob/dc55d835d1b90c9dd5dba48d0951b8ae682e857a/R/get.R#L623-L679

Both this and #448 imply that the list returned should be an S3 class.

toph-allen:
## Intent

Allow the list of content returned by `search_content()` to be converted to a data frame.

Fixes #451 

## Approach

- Add a class name to the list returned by `search_content()`
- Provide `as_tibble()` and `as.data.frame()` methods for that function, exactly following the pattern used in `integrations.R`.

## Checklist

- [ ] Does this change update `NEWS.md` (referencing the connected issue if necessary)? [not needed]

### DIFF
--- a/man/bundle_dir.Rd
+++ b/man/bundle_dir.Rd
@@ -21,7 +21,7 @@ Bundle A bundle object
 Creates a bundle from a target directory.
 }
 \examples{
-\dontshow{if (identical(Sys.getenv("IN_PKGDOWN"), "true")) (if (getRversion() >= "3.4") withAutoprint else force)(\{ # examplesIf}
+\dontshow{if (identical(Sys.getenv("IN_PKGDOWN"), "true")) withAutoprint(\{ # examplesIf}
 
 bundle_dir(system.file("tests/testthat/examples/shiny/", package = "connectapi"))
 \dontshow{\}) # examplesIf}
--- a/man/bundle_path.Rd
+++ b/man/bundle_path.Rd
@@ -16,7 +16,7 @@ Bundle A bundle object
 Define a bundle from a path (a path directly to a tar.gz file)
 }
 \examples{
-\dontshow{if (identical(Sys.getenv("IN_PKGDOWN"), "true")) (if (getRversion() >= "3.4") withAutoprint else force)(\{ # examplesIf}
+\dontshow{if (identical(Sys.getenv("IN_PKGDOWN"), "true")) withAutoprint(\{ # examplesIf}
 
 bundle_path(system.file("tests/testthat/examples/static.tar.gz", package = "connectapi"))
 \dontshow{\}) # examplesIf}
--- a/man/bundle_static.Rd
+++ b/man/bundle_static.Rd
@@ -26,7 +26,7 @@ directory, generates a basic manifest file (using the first file as the
 NOTE: the \code{rsconnect} package is required for this function to work properly.
 }
 \examples{
-\dontshow{if (identical(Sys.getenv("IN_PKGDOWN"), "true")) (if (getRversion() >= "3.4") withAutoprint else force)(\{ # examplesIf}
+\dontshow{if (identical(Sys.getenv("IN_PKGDOWN"), "true")) withAutoprint(\{ # examplesIf}
 
 bundle_static(system.file("logo.png", package = "connectapi"))
 \dontshow{\}) # examplesIf}
--- a/man/connect.Rd
+++ b/man/connect.Rd
@@ -75,7 +75,7 @@ fallback_key <- Sys.getenv("VIEWER_ROLE_API_KEY")
 client <- connect(token = token, token_local_testing_key = fallback_key)
 }
 
-\dontshow{if (identical(Sys.getenv("IN_PKGDOWN"), "true")) (if (getRversion() >= "3.4") withAutoprint else force)(\{ # examplesIf}
+\dontshow{if (identical(Sys.getenv("IN_PKGDOWN"), "true")) withAutoprint(\{ # examplesIf}
 
 # default is to read CONNECT_SERVER and CONNECT_API_KEY environment variables
 # this example will read TEST_1_SERVER and TEST_1_API_KEY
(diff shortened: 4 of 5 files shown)


DESCRIPTION
```diff
@@ -64,7 +64,7 @@ VignetteBuilder:
     knitr
 Encoding: UTF-8
 Language: en-US
-RoxygenNote: 7.3.2
+RoxygenNote: 7.3.3
```
jonkeane: I feel like I've seen this in a bunch of PRs, maybe it's "just" that the others haven't merged yet, but IMO it's nicer for the reviewer if this kind of spurious change isn't included in a bunch of PRs. Do it once, get that merged and keep it out of your others PRs.

R/content.R
```diff
@@ -1548,9 +1548,12 @@ search_content <- function(
     limit = limit
   )
 
-  purrr::map(res, function(x) {
+  content_list <- purrr::map(res, function(x) {
     Content$new(client, x)
   })
+
+  class(content_list) <- c("connect_content_list", class(content_list))
+  content_list
```
jonkeane: Do we need to preface this class with `connect_`? That seems like unnecessary typing for us, and if we want to communicate this is from Connect, we can do that in the print method(s) on this object. Our R6 classes are _not_ for example all prefaced with `Connect_` 

I hate `content`, as always, but I don't think that's solvable here, but so much do I wish we had a better label for this.

DESCRIPTION
toph-allen: This happens automatically when running `devtools::document()` with the latest `roxygen2` version, and since there are multiple PRs happening concurrently against `main`, it's happening in a few different places right now.

This'll _should_ stop appearing soon (once it's on `main`). I'm a little surprised by its presence here, though — I thought it was already on `main` — so I'll make sure I look into it.

R/content.R
toph-allen: The S3 classes we've been playing with so far (just integrations and @karawoo's work on users right now) all use a `connect_` prefix.

My thinking is that S3 classes are more common than R6 classes and are a lot less strict in their form, so having more verbosely & explicitly-named classes here is better.

There wasn't much _discussion_ of this approach at the time, so I didn't think it was controversial! (integrations: #431, #437, #440, user: #462)

If we conclude that it should just be named `content_list`, I feel like I should also change the name of the integrations classes.

R/content.R
```diff
@@ -1573,3 +1576,45 @@ search_content <- function(
...
+#' Converts a list returned by [search_content()] to a tibble.
+#'
+#' @param x A `connect_content_list` object.
+#' @param ... Unused.
+#'
+#' @return A tibble with one row per content item.
+#' @export
+as_tibble.connect_content_list <- function(x, ...) {
+  content_data <- purrr::map(x, "content")
+  parse_connectapi_typed(content_data, connectapi_ptypes$content)
```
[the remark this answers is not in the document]
toph-allen: I'm not sure. 😣 I know you want me to remove — _I_ would keep it, personally!

I'll try parsing it un-typed and see how that feels.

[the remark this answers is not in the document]
toph-allen: Parsing without providing a prototype (using `parse_connectapi`) still attempts to parse columns based on their contents, so we still get columns with e.g. logical, numeric, character. But those could technically be inconsistent if the lists are short.

Date times parse as character, which would shift the burden onto our customers to figure out the correct way to parse. I'm not sure if [the logic we have about parsing date times](https://github.com/posit-dev/connectapi/blob/a4db014da539b64c3f6d2fe663b4d2de031f4a1c/R/parse.R#L124-L136) is necessary complexity or not, but if it is, then it seems like it adds value and utility to handle that for the user.

Perhaps I'm thinking about this wrong though — perhaps we should just parse _everything_ as a character vector — is that what you're thinking?

R/content.R
```diff
@@ -1573,3 +1576,45 @@ search_content <- function(
...
+#' Converts a list returned by [search_content()] into a data frame.
+#'
+#' @param x A `connect_content_list` object (from [search_content()]).
+#' @param row.names Passed to [base::as.data.frame()].
+#' @param optional Passed to [base::as.data.frame()].
+#' @param ... Passed to [base::as.data.frame()].
+#'
+#' @return A `data.frame` with one row per content item.
+#' @export
+as.data.frame.connect_content_list <- function(
```
nealrichardson: How would this method differ for any other list-of-records from the API? This makes me think you want to have some inheritance, set the S3 class as something like `c("connect_content_list", "connect_list")`, and then you define the method on connect_list so any list-type container would be supported.

DESCRIPTION
toph-allen: I reverted this and put it in its own separate PR.

R/content.R
```diff
@@ -1573,3 +1576,45 @@ search_content <- function(
...
+  x,
+  row.names = NULL, # nolint
+  optional = FALSE,
+  ...
+) {
+  content_tbl <- as_tibble(x)
```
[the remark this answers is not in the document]
toph-allen: Yeah, it does seem backwards. This is because `parse_connectapi_typed` produces tibbles by default. I believe there was discussion of this in an integrations PR, lemme find it.

Yep: https://github.com/posit-dev/connectapi/pull/426#discussion_r2236349002 :)

R/content.R
```diff
@@ -1548,9 +1548,12 @@ search_content <- function(
     limit = limit
   )
 
-  purrr::map(res, function(x) {
+  content_list <- purrr::map(res, function(x) {
```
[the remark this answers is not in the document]
nealrichardson: IME the lapply functions work by `[[`ing elements so yes, it would work--though that's an empirical question here.

> Would defining a [[ method just shift that latency to other calls?

It defers it, yes, but the point is that in many cases, you don't ever need to make the R6 object, so you never pay that penalty. To make a data.frame as you are doing here, for example, you are making 7000 R6 objects and then immediately iterating over them to extract the data back out of it.

PR #475, opened 2025-10-25
ci: Use `with-connect` action for integration tests

### LINKED ISSUE #231: Simplify docker compose for test setup
https://github.com/rstudio/connectapi/pull/228/files#r1561773502

Also clean up the env var/secret names

### LINKED ISSUE #240: Test against dev Connect image
#178 started trying to set up a nightly job for this. The image names are different now though. 

Also to consider: the right place for this might be in the Connect product's test matrix, not here.

### LINKED ISSUE #446: Tidy up Makefile, docker compose files, etc.
The Makefile contains targets whose names are unclear, possibly duplicated, or no longer used. None of them are used by CI.

- `connect-up` / `connect-down`
  - Adds containers to a network (`.github/local/make-network.yml`)
  - Uses `inst/ci/test-connect-lic.yml`, which is the CI compose file
- `test-env-up` / `test-env-down`
  - Uses `.github/local/test-connect-ci.yml` (which is _not_ the CI compose file)
  - Also uses `make-network.yml`
- `mail-up` / `mail-down`
  - Don't seem to be used anywhere?
  - Use `.github/local/mail.yml`
- `test` / `test-run` / `test-run-i`
  - Seem to reference an old & broken workflow, can probably be deleted.
  - Reference `test.yml`

Compose files

- `inst/ci/test-connect-lic.yml` creates a single service named `connect` with `scale: 2`. Used in CI.
- `.github/local/test-connect-ci.yml` creates two services, `connect1` vs `connect2`. Not used in CI.
- `.github/local/mail.yml`
- `test.yml` — Creates a container that runs a file `/test/inst/ci/run_tests.R` that doesn't exist (seems to have been moved to `.github/local/run_tests.R`.

nealrichardson:
## Intent

Remove `HackyConnect` and the related docker compose tooling for running integration tests, in favor of a new `with-connect` utility that @tdstein started work on.

The main purpose for doing this here is to prove out how `with-connect` and its accompanying GitHub Action can be used in CI, so that we can roll it out to the range of other open-source packages that need it. The code deletion here is a nice benefit.

In the process, I discovered that the v1 git endpoints weren't operational in 2022.10.0, as we thought in #459, but rather in 2022.12.0, so I updated those version checks.

Fixes #446
Fixes #240 (as best we can)
Fixes #231 (in a way)

## Approach

Iteration back and forth here and on https://github.com/nealrichardson/with-connect/tree/dev, my fork of the action. Once everything was switched over and confirmed working, delete as much as I could, including some stale local dev setup and a test file that was all skips.

There is some further simplification of the integration test suite to do, but I've deferred that here. 

## Checklist

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -51,8 +51,8 @@ jobs:
         env:
           CONNECTAPI_INTEGRATED: "true"
         with:
-          connect-version: ${{ matrix.version }}
-          connect-license: ${{ secrets.CONNECT_LICENSE_FILE }}
+          version: ${{ matrix.version }}
+          license: ${{ secrets.CONNECT_LICENSE_FILE }}
           # TODO: rewrite tests to use CONNECT_* env vars directly
           command: |
             Rscript -e '
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -37,7 +37,7 @@ jobs:
         uses: nealrichardson/with-connect@dev
         with:
           # Runs on the default version in the with-connect action
-          connect-license: ${{ secrets.CONNECT_LICENSE_FILE }}
+          license: ${{ secrets.CONNECT_LICENSE_FILE }}
           # TODO: rewrite tests to use CONNECT_* env vars directly
           command: |
             Rscript -e '
--- a/.github/workflows/test-coverage.yaml
+++ b/.github/workflows/test-coverage.yaml
@@ -34,7 +34,7 @@ jobs:
           CONNECTAPI_INTEGRATED: "true"
         with:
           # Runs on the default version in the with-connect action
-          connect-license: ${{ secrets.CONNECT_LICENSE_FILE }}
+          license: ${{ secrets.CONNECT_LICENSE_FILE }}
           # TODO: rewrite tests to use CONNECT_* env vars directly
           command: |
             Rscript -e '


.github/workflows/integration-tests.yaml
```diff
@@ -15,35 +15,28 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - "2025.09.0" # jammy
```
nealrichardson: I *think* we can also add `"latest"` here and that will solve #240, if we want.

.github/workflows/integration-tests.yaml
```diff
@@ -52,24 +45,17 @@ jobs:
...
         with:
-          name: ${{ runner.os }}-${{ matrix.version }}-results
-          path: check
+          connect-version: ${{ matrix.version }}
+          connect-license: ${{ secrets.CONNECT_LICENSE_FILE }}
+          # TODO: rewrite tests to use CONNECT_* env vars directly
+          command: |
+            Rscript -e '
+              Sys.setenv(
+                CONNECTAPI_INTEGRATED="true",
```
nealrichardson: It's possible that this could go in the `env` of the action step, will try that. It's not possible to have an argument in `with` that is a YAML mapping, so `env` can't go there.

.github/workflows/pr-commands.yaml
```diff
@@ -7,55 +7,6 @@ on:
 name: Commands
 
 jobs:
-  update_snapshot:
```
nealrichardson: I don't think this is ever used.

tdstein (commented):
Hooray! This is awesome!

PR #484, opened 2025-12-10
ci: Integration test against nightly docker image

nealrichardson:
## Intent

See https://github.com/posit-dev/with-connect/pull/37. This will run integration tests against a nightly build of Connect.


tdstein:
Approved since the linter error is unrelated to this change.

PR #487, opened 2025-12-19
Fix CI

### LINKED ISSUE #485: ci: linting fails since lintr 3.3.0 release
Specifically, the first point here: https://github.com/r-lib/lintr/blob/main/NEWS.md#deprecations--breaking-changes-1

On CRAN since November 27, but I guess we haven't had a PR go through since then? Observed failure on a change that does not touch R code [here](https://github.com/posit-dev/connectapi/actions/runs/20117418256/job/57730059173?pr=484).

We should either update to use the `|>` or skip the pipe linter rule.

karawoo:
## Intent

Explicitly sets the pipe linter to `"auto"` instead of the new default `"|>"`.  Also removes the cyclocomp linter line; it's no longer necessary as it's not a lintr default anymore since 3.2.0.

Fixes #485 

## Approach

I chose to keep our existing magrittr pipe code instead of updating to the base pipe, mainly because we don't currently set a minimum R version in the DESCRIPTION. Using the base R pipe would require us to set the minimum to 4.1. This might be okay, and I don't feel too strongly either way, but this at least gets CI passing again.

## Checklist

### DIFF
--- a/tests/integrated/helper.R
+++ b/tests/integrated/helper.R
@@ -9,7 +9,8 @@ expect_ptype_equal <- function(actual, expected, exact = TRUE) {
 }
 
 skip_if_connect_older_than <- function(client, version) {
-  if (numeric_version(safe_server_version(client)) < numeric_version(version)) {
+  current <- numeric_version(simplify_version(safe_server_version(client)))
+  if (current < numeric_version(version)) {
     skip(paste("Requires Connect >=", version))
   }
 }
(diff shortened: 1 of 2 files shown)


karawoo:
> Odd that the preview integration tests didn't fail before, but thanks for the clean fix!

AFAICT from the logs on https://github.com/posit-dev/connectapi/pull/484, the tests never actually ran on that PR. With the changes in https://github.com/posit-dev/with-connect/pull/38 they run and fail. But fixed now.

PR #491, opened 2025-12-31
chore: remove `get_job()`

### LINKED ISSUE #490: refactor: use v1 job API, add function to get logs
#343 solved `get_jobs()` but not `get_job()`, which still calls the unversioned endpoint. It looks like the data munging steps that were added in `get_jobs()` could be factored out and used in `get_job()`, but there another difference that I can see in the fixtures that were recored in #343: the unversioned job endpoint includes logs, but the v1 does not. There is a separate endpoint for that in v1, and it looks like it returns data in a different structure too: https://docs.posit.co/connect/api/#get-/v1/content/-guid-/jobs/-key-/log

Since the `get_job()` function was marked as experimental, I don't think that it's worth trying to mimic the shape of the data that the old API returned. We should probably just break compat and make a new `get_logs()` function--that's probably the reason you would have called `get_job()` anyway?

toph-allen:
## Intent

Remove `get_job()`, which called an unversioned endpoint.

Fixes #490

## Approach

The function had been deprecated for a while, and has replacements.

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,8 +4,9 @@
 
 ## Breaking changes
 
-- Removed `get_job()`, which was deprecated in 0.6.0. Use `get_jobs()`,
-  `get_job_list()`, or `get_log()` instead. (#491)
+- Removed `get_job()`, which was deprecated in 0.6.0. Instead, use `get_jobs()`
+  to get a data frame of jobs, use `get_job_list()` to get a list of job
+  objects, and use `get_log()` to get the log for a job (#491).
 
 # connectapi 0.9.0
 


NEWS.md
```diff
@@ -2,6 +2,11 @@
 
 - Removed `prefix` argument to `connect()` function, which was a convenience utility for testing. (#477)
 
+## Breaking changes
+
+- Removed `get_job()`, which was deprecated in 0.6.0. Use `get_jobs()`,
+  `get_job_list()`, or `get_log()` instead. (#491)
```
nealrichardson: Is it worth briefly noting why you would use one of these three alternatives? Like, to see the logs, call `get_log()`, or to see other metadata about the job, use the others?

PR #493, opened 2026-01-12
chore: remove swap_vanity_url() and accommodate new API fields

karawoo:
## Intent

#491 removed `get_job()` which was deprecated in 0.6.0. Since we are getting ready to do another connectapi release I think it makes sense to remove the other function that was deprecated in 0.6.0, `swap_vanity_url()`. 

## Checklist

### DIFF
--- a/R/ptype.R
+++ b/R/ptype.R
@@ -20,7 +20,8 @@ connectapi_ptypes <- list(
   groups = tibble::tibble(
     "guid" = NA_character_,
     "name" = NA_character_,
-    "owner_guid" = NA_character_
+    "owner_guid" = NA_character_,
+    "gid" = NA_character_
   ),
   usage_shiny = tibble::tibble(
     "content_guid" = NA_character_,
--- a/tests/integrated/test-lazy.R
+++ b/tests/integrated/test-lazy.R
@@ -107,7 +107,7 @@ test_that("groups works", {
   expect_type(colnames(groups_list), "character")
   expect_gt(length(colnames(groups_list)), 1)
 
-  expect_ptype_equal(groups_list_local, connectapi_ptypes$groups)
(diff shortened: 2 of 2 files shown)


R/ptype.R
```diff
@@ -20,7 +20,8 @@ connectapi_ptypes <- list(
   groups = tibble::tibble(
     "guid" = NA_character_,
     "name" = NA_character_,
-    "owner_guid" = NA_character_
+    "owner_guid" = NA_character_,
+    "gid" = NA_character_
```
jonkeane: Is this separate from removing the swap vanity url stuff? I'm fine to bundle it in one PR, but I want to make sure I'm following.

tests/integrated/test-lazy.R
```diff
@@ -107,7 +107,7 @@ test_that("groups works", {
   expect_type(colnames(groups_list), "character")
   expect_gt(length(colnames(groups_list)), 1)
 
-  expect_ptype_equal(groups_list_local, connectapi_ptypes$groups)
+  expect_ptype_equal(groups_list_local, connectapi_ptypes$groups, exact = FALSE)
```
jonkeane: Same with this, this looks like it's something slightly different from the swap vanity urls thing?

R/ptype.R
karawoo: yeah it was a change to fix integration tests in CI on the preview version of Connect. I can also pull it out into a separate PR too if you prefer; either way I need to figure out what is going on with the integration tests for 2022-2023 versions first.

karawoo:
> Also the integration tests look like a licensing issue, maybe we need to reissue those? 

Possibly but it's weird to me that integration tests on other versions of connect that use the same license file are passing

PR #496, opened 2026-02-06
remove map_dfr - fixes posit-dev/connectapi#422

AntLP:
## Intent

As described in the issue, some functions had a dependency on `dplyr` through the use of `purrr::map_df` and `purrr::map_dfr`.

Fixes posit-dev/connectapi#422

## Approach

Replaced the calls to `purrr::map_df` and `purrr::map_dfr` with a combination of `purrr::map` and `purrr::list_rbind`, which removes the dependency on `dplyr`.

I did not remove the `purrr::map_dfr` usages in the tests, `dplyr` is therefore still required to run the tests.

## Checklist

### DIFF
--- a/R/get.R
+++ b/R/get.R
@@ -1056,11 +1056,7 @@ get_aws_credentials <- function(connect, user_session_token, audience = NULL) {
 #' }
 #'
 #' @export
-get_aws_content_credentials <- function(
-  connect,
-  content_session_token = NULL,
-  audience = NULL
-) {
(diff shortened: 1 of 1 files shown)


karawoo:
Thank you for this contribution! The failing Ci issues all appear to be due to repository secrets not being available to forks; I've tried your changes locally and everything looks good. I'll merge this and if any CI stuff continues to fail I'll fix that on my own as a followup.

PR #505, opened 2026-02-24
grow results by page in page_cursor

### LINKED ISSUE #501: Improve performance in `page_cursor`
`page_cursor` iteratively grows the results list which can be very slow if the number of pages is very large

https://github.com/posit-dev/connectapi/blob/532576024b3943667b00a488bea6706ea888f010/R/page.R#L34

We cannot instantiate `res` with the exact correct size in advance, which would be the best solution, because we do not know the full number of pages. However we can and should still improve this.

karawoo:
## Intent

Fixes #501

## Approach

We were appending results iteratively with `c()` and, while not the slowest part of this code, it was improved by growing the list page by page and then combining once at the end with `do.call()`. Unfortunately we cannot pre-allocate the list because the API does not return the total number of pages, we have to go through them one by one until we reach the end.

For a `get_usage_static()` call returning 175k results this shaved off about half a second (~11.5 seconds to just under 11). Not a huge improvement but still worth making since the code is no more complex.

While I was in there I also noticed some incorrect documentation and unnecessary tidy evaluation that I've also removed.

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@
   removed in a future release. (#500)
 - Added a single retry to `content_restart()` to more robustly clean up
   temporary environment variables. (#498)
+- Improved performance of `page_cursor()`. (#501)
   
 ## Breaking changes
 


jonkeane:
One comment: given [your comment](https://github.com/posit-dev/connectapi/issues/501#issuecomment-3949318321) on #501 do we want to say that this PR fixes #501? Or just that it helps?

PR #508, opened 2026-02-25
Add `content_guid` filtering to `get_usage()`

jonkeane:
## Intent

Now that connect accepts GUID filters on the hits endpoint, add that to the sdk

## Checklist


karawoo:
@jonkeane actually if you don't mind, i'll just do it -- it'll be a quick fix and will unblock me moving forward with the cran release

PR #511, opened 2026-03-04
Remove strict type checking

jonkeane:
## Intent

Remove the strict type checking. This is basically a oneshot from claude. Putting up here to see if CI passes + what folks think

## Approach

Only to coercion around date times since we get RFC 3339 back which are not auto-converted to `posix.ct` (though maybe we don't even need to do that, just accept these are characters and then only at presentation / dataframing do we make them a native type.

## Checklist


karawoo:
I have a WIP branch to do something similar that I hadn't gotten opened yet, let me open a PR to compare. I think there are some pros/cons of each

PR #513, opened 2026-03-05
Improve response parsing: no ptypes, faster datetimes

### LINKED ISSUE #483: Remove or drastically overhaul type parsing even when producing data frames
Refactor `connectapi`'s type parsing logic, specifically for functions like get_content() and get_usage(), as it's very slow with large objects.

My hunch is that `parse_connectapi` iterates over the JSON as a list (i.e. row-wise), and tries to preserve nested JSON as list-columns. Experimentally, just converting everything with `jsonlite::fromJSON(flatten = TRUE)` speeds up e.g., get_content() from ~10.5s to ~3.5s) on Dogfood.

It would probably be much faster to just flatten and convert to a character data frame. I *think* it's important to retain date parsing, and perhaps specific columns should be parsed as integers and character vectors, but I'm not sure what the best way to manage that is (we'd also like to remove the need to maintain a list of ptypes).

Closely related to #473 — that talks about removing the parsing when creating a list of objects, but this talks about existing data.frame-returning functions.

karawoo:
## Intent

Fixes #483

## Approach

Removes the `connectapi_ptypes` dictionary and vctrs-based type coercion system (`ensure_columns`, `ensure_column`, `vec_cast`). Instead, each getter function declares its own `datetime_cols` and applies lightweight post-parse coercion via `coerce_datetime()`, which handles RFC 3339 strings, epoch seconds, POSIXct pass-through, and all-NA columns.     

Parsing pipeline:
- `Connect$request()` now uses `jsonlite::fromJSON()` instead of `httr::content(as = "parsed")`, giving us control over jsonlite's simplification behavior
- `parse_connect_rfc3339()` uses a vectorized substr-based parser (faster than strptime on large vectors)
- `page_cursor()` fetches subsequent pages with simplify=TRUE so jsonlite builds data frames in C, then combines with `vctrs::vec_rbind()`

I started this work while trying to improve performance of `get_usage_static()` (see profiling on https://github.com/posit-dev/connectapi/issues/501#issuecomment-3949318321). With this branch, `get_usage_static()` is ~20% faster than on main for 175k records.

## Checklist


R/content.R
```diff
@@ -832,22 +848,19 @@ terminate_jobs <- function(content, keys = NULL) {
...
-      strict = TRUE
-    )
-  )
+  res_df <- parse_connectapi_typed(res_content)
   # Errors will not have the job_key.
   res_df$job_key <- keys
-  res_df
+  # Keep only the columns relevant to job termination; the API response
+  # includes extra fields (e.g. payload, guid) on error that vary by outcome.
+  keep <- c("app_id", "app_guid", "job_key", "job_id", "result", "code", "error")
```
jonkeane: Is it a problem that we get variable data at this point?

karawoo: I believe that the variable data is for error cases (i.e. trying to terminate a job that is not currently active, which returns a 409 but doesn't raise an R error) vs. successful requests. I guess it's debatable whether we should be raising an R error more eagerly, but I think the behavior of accommodating different field names is consistent with main.

The one difference here is that main will always return the columns of `keep` whereas on this branch if any columns from `keep` are missing across all responses, they'd be omitted. I'll update to make it more consistent with main.

karawoo: Gotcha yeah, if we were returning a list then I agree it'd make more sense to not worry about the columns -- the responses are what they are, and that may include different fields.

R/content.R
```diff
@@ -788,7 +788,23 @@ get_jobs <- function(content) {
...
-  parse_connectapi_typed(jobs, connectapi_ptypes$jobs, strict = TRUE)
+  out <- parse_connectapi_typed(
+    jobs,
+    datetime_cols = c("start_time", "end_time", "last_heartbeat_time", "queued_time")
+  )
+
+  # The older /applications/ endpoint returns timestamps as Unix epoch integers
+  # and ID fields as integers. Normalize to match the v1 endpoint's types.
+  # For the v1 endpoint these are already character/POSIXct, so the coercions
+  # are no-ops.
```
jonkeane, line 799: It's probably not worth too much digging, but if you know off the top of your head: how old are older versions here?

R/content.R
```diff
@@ -1141,7 +1154,8 @@ get_bundles <- function(content) {
   validate_R6_class(content, "Content")
   bundles <- content$get_bundles()
 
-  parse_connectapi_typed(bundles, connectapi_ptypes$bundles)
+  out <- parse_connectapi_typed(bundles, datetime_cols = "created_time")
+  coerce_fs_bytes(out, "size")
```
jonkeane, line 1158: We can keep this here, but I do wonder if providing these as `fs::byte` types is really useful?

tests/integrated/test-lazy.R
```diff
@@ -66,9 +64,13 @@ test_that("usage_shiny works", {
...
-  expect_gt(length(colnames(shiny_usage)), 1)
 
-  expect_ptype_equal(shiny_usage_local, connectapi_ptypes$usage_shiny)
+  # No shiny apps are deployed in integration tests, so this may be empty.
+  if (nrow(shiny_usage_local) > 0) {
+    expect_gt(length(colnames(shiny_usage)), 1)
+    expect_true("content_guid" %in% names(shiny_usage_local))
+    expect_s3_class(shiny_usage_local$started, "POSIXct")
+  }
 })
```
jonkeane, line 74: I assume the `if` here is because a 0-row df doesn't get coerced to this type? But is that true / is that what we want?

tests/testthat/test-get.R
```diff
@@ -404,12 +407,12 @@ with_mock_dir("2025.04.0", {
...
     )
 
-    # Check conversion to data.frame
+    # Check conversion to data.frame (with unnesting)
     usage_df <- as.data.frame(usage)
     expect_equal(
       usage_df,
       data.frame(
-        id = c("8966707", "8966708", "8967206", "8967210", "8966214"),
+        id = c(8966707L, 8966708L, 8967206L, 8967210L, 8966214L),
```
jonkeane: Are these supposed to remain characters? I believe they should (given how we do our IDs), but maybe I'm missing something?

jonkeane (commented):
Sorry I commented outside of a review so that I didn't accidentally leave them without sending. Overall this looks good, and I'm pro getting this on main and using it locally to confirm it works well. 

Do you think it would be a good thing to add a test or two that simulates the changes that we had to do with #512 ?

tests/testthat/test-get.R
karawoo: they changed because now we return whatever the API returns rather than trying to coerce to specific types, and the test fixtures have integers rather than characters. I'll update those but the behavior here is correct, and only seems confusing because these are unit tests that can get out of sync with connect's actual response types.

tests/integrated/test-lazy.R
karawoo: if there are no shiny apps then the usage data returns an empty response with no rows OR columns, so we can't assert the column names/types

jonkeane: Ah, hmmm that seems not ideal, but if it's not a regression in this work that's fine. But maybe a follow up would be good?

karawoo: It _is_ a regression in this work. Previously we would have been able to construct a 0-row data frame with
the expected columns based on the ptypes. Since we want to remove the ptypes, we now have no way to know what columns to create. The regression here is inherent to the motivation of the PR. I'm not sure what follow-up we could do that wouldn't bring back ptypes in some form.

R/content.R
karawoo: `GET /v1/content/{guid}/jobs` was added in October 2022 https://docs.posit.co/connect/news/#rstudio-connect-2022.10.0; the old applications endpoint was removed July 2025.

tests/integrated/test-lazy.R
jonkeane: Aaah ok, I get it now: the response from the server is a different shape if there are no visits to shiny apps to report, yeah?

I did a bit of poking and it looks like that is also common across other endpoints too (e.g. content search returns an empty list when there are no matches). So that will have a similar issue too.

karawoo:
@jonkeane how would you feel about re-reviewing this so we can merge and tackling the question of whether to construct the expected API fields based on the swagger schema separately?

R/page.R
```diff
@@ -24,20 +24,34 @@ page_cursor <- function(client, req, limit = Inf) {
...
-  # collect whole pages, then flatten once at the end
-  pages <- list(response$results)
-  n_items <- length(response$results)
+  # Convert the first page (list-of-lists from simplify=FALSE) to a data frame.
+  # Subsequent pages use simplify=TRUE so jsonlite builds data frames in C,
+  # which is significantly faster for high-volume endpoints.
+  first_results <- response$results
+  if (length(first_results) > 0 && !is.data.frame(first_results)) {
+    first_results <- parse_connectapi(first_results)
+  }
```
jonkeane, line 33: Ah nice. I would love to know how much this speeds up, do you happen to have even rough measures?

R/parse.R
```diff
@@ -165,49 +158,41 @@ coerce_datetime <- function(x, to, ...) {
...
-  if (is.integer(x)) {
-    x <- as.double(x)
-  }
-  stopifnot(is.double(x))
-  stopifnot(is.character(tzone))
-  structure(x, tzone = tzone, class = c("POSIXct", "POSIXt"))
+  # Build epoch seconds directly: date days * 86400 + time of day - tz offset
+  epoch <- as.double(dates) * 86400 + hour * 3600 + min * 60 + sec - tz_offset
+  result[!na_mask] <- epoch
+  .POSIXct(result, tz = Sys.timezone())
```
jonkeane: This change seems a little more fragile than what as there before. If I'm reading the diff correctly (and I might not be!) the core bit is:

```
 format_string <- "%Y-%m-%dT%H:%M:%OS%z"
  as.POSIXct(x, format = format_string, tz = Sys.timezone())
```

Is it possible for us to keep that and not do the pulling-values-out-of-the string bit (possibly with adding the `NA` masking bits?).  That reads to me as much more direct + clear what's going on?

tests/testthat/test-get.R
```diff
@@ -471,6 +473,35 @@ with_mock_dir("2025.04.0", {
...
+    usage <- get_usage(
+      client,
+      from = as.POSIXct("2025-04-01 00:00:01", tz = "UTC")
+    )
+    usage_df <- as.data.frame(usage)
+    expect_type(usage_df$id, "character")
+    expect_equal(usage_df$id, c("8966707", "8966708", "8967206", "8967210", "8966214"))
+    expect_s3_class(usage_df$timestamp, "POSIXct")
+  })
+})
```
jonkeane, line 503: Thanks for this test!

jonkeane (approved):
> @jonkeane how would you feel about re-reviewing this so we can merge and tackling the question of whether to construct the expected API fields based on the swagger schema separately?

Agreed! I've approved, though I do think we might want to make the timestamp parser a little more readable before we merge. Otherwise, the tests look good, and if they are passing, I think we are good to merge and iterate

R/parse.R
karawoo: yeah, this was a performance optimization but I agree the new code is a lot more complex and harder to read. I'll simplify it. for pulling ~175k records via `get_usage_static(limit = Inf)`, the difference your suggestion and how it's currently written is about half a second (5.5 seconds vs. 6) and still about a second faster than on main.

jonkeane: > the difference your suggestion and how it's currently written is about half a second (5.5 seconds vs. 6)

*and* that half a second is less than 10% even!

PR #521, opened 2026-03-23
feat: read session token from file, if present

nealrichardson:
## Intent

Closes #518

## Approach

Write a failing test, make it pass, update the docs

## Checklist

- [ ] Does this change resolve integration test failures for a preview build of Connect? If so (and if you are on the Connect team), please be sure to update the Connect changelog to reflect that users will need to upgrade their version of `connectapi`.

### DIFF
--- a/R/get.R
+++ b/R/get.R
@@ -871,8 +871,10 @@ get_oauth_credentials <- function(
 #' @param content_session_token Optional. The content session token. This token
 #' can only be obtained when the content is running on a Connect server. The
 #' token identifies the service account integration previously configured by
-#' the publisher on the Connect server. Defaults to the value from the
-#' environment variable: `CONNECT_CONTENT_SESSION_TOKEN`
+#' the publisher on the Connect server. Defaults to the value found in the file
+#' indicated by the `CONNECT_CONTENT_SESSION_TOKEN_FILE` environment variable
+#' on Connect >= 2026.02.0, or from the value of the environment variable
+#' `CONNECT_CONTENT_SESSION_TOKEN` on earlier versions.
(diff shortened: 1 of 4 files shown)


R/get.R
```diff
@@ -918,6 +920,14 @@ get_oauth_content_credentials <- function(
     error_if_less_than(connect$version, "2025.07.0")
   }
 
+  # First, try CONNECT_CONTENT_SESSION_TOKEN_FILE
+  if (is.null(content_session_token)) {
+    token_file <- Sys.getenv("CONNECT_CONTENT_SESSION_TOKEN_FILE")
+    if (nzchar(token_file)) {
+      content_session_token <- readLines(token_file, n = 1, warn = FALSE)
```
tdstein, line 927: I don't know R primitives very well. Does the `nzchar` equality check handle conditions where the file specified at `CONNECT_CONTENT_SESSION_TOKEN_FILE` is missing? Also, will `if (is.null(content_session_token))` evaluate to false if the file is empty?

nealrichardson: `nzchar` is handling "is the env var set and not empty". If the file it points to does not exist, `readLines()` will throw an error. If the file exists but is empty, that would return a 0-length string, which would be an invalid token but not `NULL` so it wouldn't fall back to checking the `CONNECT_CONTENT_SESSION_TOKEN` env var.

We could be more defensive around those, I suppose, but what other than raising a different error message would we do? If the session token file is invalid, why should we expect anything else to work?

tdstein: yeah yeah, I just wanted to make sure that it handled empty files in the same way as the Python implementation for consistency.

tdstein (approved):

PR #522, opened 2026-04-09
feat: support POSIT_PRODUCT env var and Connect Cloud for OAuth apps

samperman:
## Intent

Enable this library to work inside Posit Connect Cloud

## Approach

Add POSIT_PRODUCT as an alternative to RSTUDIO_PRODUCT for detecting Connect, and add on_connect_cloud() to support apps running in Connect Cloud (POSIT_PRODUCT="CONNECT_CLOUD").

## Checklist

- [ ] Does this change resolve integration test failures for a preview build of Connect? If so (and if you are on the Connect team), please be sure to update the Connect changelog to reflect that users will need to upgrade their version of `connectapi`.

### DIFF
--- a/R/connect.R
+++ b/R/connect.R
@@ -1018,7 +1018,7 @@ connect <- function(
     } else {
       con <- connect(server = server, api_key = token_local_testing_key)
       message(paste0(
-        "Called with `token` but not running on Connect. ",
+        "Called with `token` but not running on Connect or Connect Cloud. ",
         "Continuing with fallback API key."
       ))
     }
--- a/R/connect.R
+++ b/R/connect.R
@@ -1007,7 +1007,7 @@ connect <- function(
 
   if (!missing(token)) {
     error_if_less_than(con$version, "2025.01.0")
-    if (on_connect()) {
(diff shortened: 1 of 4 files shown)


mbaynton (approved):

R/connect.R
```diff
@@ -1007,7 +1007,7 @@ connect <- function(
 
   if (!missing(token)) {
     error_if_less_than(con$version, "2025.01.0")
```
karawoo, line 1009: just double checking that connect cloud uses the same versioning scheme and that this check will work appropriately there?

R/connect.R
karawoo: maybe update this message to include Connect Cloud?

karawoo (commented):
Couple of questions below, could you also please add a bullet to the changelog in NEWS.md?

R/connect.R
samperman: This should actually get the version of Connect under the hood. I'll verify in connect cloud before merging.

karawoo (approved):

R/connect.R
karawoo: ah ok great

PR #524, opened 2026-04-22
fix for change in repo/branches format

karawoo:
## Intent

Fix integration tests, which failed on the nightly connect build this week: 
https://github.com/posit-dev/connectapi/actions/runs/24657089648/job/72093203473

## Approach

Add a helper than can parse either the old format or the new. 

## Checklist

- [x] Does this change resolve integration test failures for a preview build of Connect? If so (and if you are on the Connect team), please be sure to update the Connect changelog to reflect that users will need to upgrade their version of `connectapi`.

### DIFF
--- a/R/git.R
+++ b/R/git.R
@@ -41,9 +41,8 @@ repo_check_branches <- function(client, repository) {
   branches_task <- Task$new(connect = client, task = branches_raw)
 
   task_res <- poll_task(branches_task, callback = NULL)
-  task_data <- task_res$get_data()
-  stopifnot(identical(task_data$type, "git-repo-ref-branch-array"))
-  branches <- purrr::map(task_data$data, ~ .x$branch)
+  branch_entries <- parse_repo_branches(task_res$get_data())
+  branches <- purrr::map(branch_entries, ~ .x$branch)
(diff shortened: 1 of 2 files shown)


karawoo:
> Does this change resolve integration test failures for a preview build of Connect? If so (and if you are on the Connect team), please be sure to update the Connect changelog to reflect that users will need to upgrade their version of connectapi.

Not sure whether this makes sense to surface in the changelog given that it deals with an internal, unversioned API. That said, if a customer is using connectapi for things dealing with git and they run into this issue, they should have some way of knowing that they need to upgrade. It's not their fault that connectapi is using internal APIs. @jonkeane let me know what you think.

jonkeane (approved):
These changes look good, thanks for jumping on these

> Not sure whether this makes sense to surface in the changelog given that it deals with an internal, unversioned API.

I agree with your assessment that it's not worth calling out in the Connect changelog.

karawoo:
Sounds good, I also checked the connect extensions and didn't find any that rely on this functionality, so I think the extensions should keep running fine regardless (breaking these was what prompted us to add that PR checklist item).

